PR #45, opened 2024-03-20
feat: connect to xmidt cluster

denopink:
WIP

Pending TODOs
- [x] add `ws` provider & config (referred as `type Client struct`)
- [x] add openfail xmidt connection
- [x] add xmidt connection with jwt


internal/websocket/options.go
```diff
@@ -54,7 +68,10 @@ func FetchURLTimeout(d time.Duration) Option {
 		func(ws *Websocket) error {
 			if d < 0 {
 				return fmt.Errorf("%w: negative FetchURLTimeout", ErrMisconfiguredWS)
+			} else if d == 0 {
+				d = fetchUrlTimeoutDefault
```
denopink: @schmidtw I added defaults in case the agent starts with only a `fetch_url` and nothing else

```yaml
client:
  fetch_url: https://localhost:8080/api/v2/device
```

cmd/xmidt-agent/config.go
```diff
@@ -26,6 +28,41 @@ type Config struct {
 	Storage          Storage
 }
 
+type Client struct {
```
schmidtw: Rename `Client` to be something like `Websocket` that way we can support this and `QUIC` in the future.

internal/websocket/ws.go
```diff
@@ -115,15 +122,15 @@ func New(opts ...Option) (*Websocket, error) {
...
-		FetchURLTimeout(30 * time.Second),
-		PingInterval(30 * time.Second),
-		PingTimeout(90 * time.Second),
-		ConnectTimeout(30 * time.Second),
-		KeepAliveInterval(30 * time.Second),
-		IdleConnTimeout(10 * time.Second),
-		TLSHandshakeTimeout(10 * time.Second),
-		ExpectContinueTimeout(1 * time.Second),
-		MaxMessageBytes(256 * 1024),
+		FetchURLTimeout(fetchUrlTimeoutDefault),
```
schmidtw: We may want to simply remove all the default values from here & instead favor moving them into the `cmd/xmidt-agent/config.go` file as built in defaults to the actual configuration.  We will also want to add an `Enabled` flag there as well ... that way through configuration this can be turned on/off & the defaults are expressed in the configuration system & to the end user.

schmidtw (approved):

PR #57, opened 2024-04-02
feat: improve ws.conn handling

denopink:
Doesn't include `func (ws *websocket) Send()` update

### DIFF
--- a/internal/websocket/e2e_test.go
+++ b/internal/websocket/e2e_test.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
 	"github.com/xmidt-org/retry"
 	"github.com/xmidt-org/wrp-go/v3"
 	ws "github.com/xmidt-org/xmidt-agent/internal/websocket"
@@ -22,11 +23,22 @@ import (
 	"nhooyr.io/websocket"
 )
 
-func TestEndToEnd(t *testing.T) {
-	var finished bool
+type EndToEndTestSuite struct {
+	suite.Suite
+	finished  bool
+	websocket *ws.Websocket
+}
 
-	assert := assert.New(t)
-	require := require.New(t)
+// Make sure that VariableThatShouldStartAtFive is set to five
+// before each test
+func (ts *EndToEndTestSuite) AfterTest(suiteName, testName string) {
+	ts.finished = true
+	ts.websocket.Stop()
+}
+
+func (ts *EndToEndTestSuite) TestEndToEnd() {
+	assert := assert.New(ts.T())
+	require := require.New(ts.T())
 
 	s := httptest.NewServer(
 		http.HandlerFunc(
@@ -47,7 +59,7 @@ func TestEndToEnd(t *testing.T) {
 
 				mt, got, err := c.Read(ctx)
 				// server will halt until the websocket closes resulting in a EOF
-				if finished {
+				if ts.finished {
 					return
 				}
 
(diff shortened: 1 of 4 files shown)


internal/websocket/e2e_test.go
```diff
@@ -44,14 +48,19 @@ func TestEndToEnd(t *testing.T) {
 				require.NoError(err)
 
 				mt, got, err := c.Read(ctx)
+				// server will halt until the websocket closes resulting in a EOF
+				if finished && errors.Is(err, io.EOF) {
```
johnabass: No need to use `errors.Is` here.  In fact, you really shouldn't.  `io.Reader.Read` is documented as returning *exactly* `io.EOF` in this case.  I realize the linter will complain, which is why I hate our linter :)

https://pkg.go.dev/io#Reader

internal/websocket/e2e_test.go
```diff
@@ -23,6 +25,8 @@ import (
 )
 
 func TestEndToEnd(t *testing.T) {
+	var finished bool
```
johnabass: Consider using a test suite, as implemented by https://pkg.go.dev/github.com/stretchr/testify/suite.  This will clean up this test code considerably, as you can move a lot of setup code into the lifecycle of the test rather than the test method.

internal/websocket/e2e_test.go
```diff
@@ -256,7 +266,7 @@ func TestEndToEndConnectionIssues(t *testing.T) {
 				require.NoError(err)
 				defer c.CloseNow()
 
-				ctx, cancel := context.WithTimeout(r.Context(), 2000000*time.Millisecond)
+				ctx, cancel := context.WithTimeout(r.Context(), 200*time.Millisecond)
```
denopink, line 283: this was a silly artifact of some test debugging code in a previous pr

internal/websocket/ws.go
```diff
@@ -152,13 +155,18 @@ func (ws *Websocket) Start() {
...
 
-	go ws.run(ctx)
+	go ws.readPump(ctx)
 }
 
 // Stop stops the websocket connection.
 func (ws *Websocket) Stop() {
 	ws.m.Lock()
+	// Avoid the overhead of the close handshake.
+	ws.conn.Close(nhws.StatusNormalClosure, "")
```
denopink, line 171: looks like nhooyr.io/websocket's websocket doesn't like it when you use its own nhws.StatusNormalClosure to close the socket... I was expecting a normal EOF: 
```
Received unexpected error:
        	            	failed to get reader: received close frame: status = StatusNormalClosure and reason = ""
```
https://github.com/xmidt-org/xmidt-agent/blob/77872d04f2fa3727349ff329eb933f608ed0a80a/internal/websocket/e2e_test.go#L48

        	Error Trace:	/Users/odc/Documents/GitHub/xmidt-org/xmidt-agent/internal/websocket/e2e_test.go:56
        	            				/Users/odc/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.21.8.darwin-amd64/src/net/http/server.go:2136
        	            				/Users/odc/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.21.8.darwin-amd64/src/net/http/server.go:2938
        	            				/Users/odc/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.21.8.darwin-amd64/src/net/http/server.go:2009
        	            				/Users/odc/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.21.8.darwin-amd64/src/runtime/asm_amd64.s:1650
        	Error:      	Received unexpected error:
        	            	failed to get reader: received close frame: status = StatusNormalClosure and reason = ""
        	Test:       	TestEndToEnd

denopink: I would assume `nhws.StatusNormalClosure` would trigger a `io.EOF` and not a unexpected error

denopink: looks like nhooyr.io/websocket's `Reader` doesn't return an `io.EOF` for a `StatusNormalClosure` (status code 1000).
https://github.com/xmidt-org/xmidt-agent/pull/57#discussion_r1548670211

just a fyi

PR #58, opened 2024-04-03
Mocktr181

piccione99:


lint.sh
```diff
@@ -0,0 +1,2 @@
+go fmt ./...
+golangci-lint run
```
denopink: gitignore?

internal/wrphandlers/mocktr181/handler.go
```diff
@@ -0,0 +1,211 @@
...
+// the handler that will be called to send the response if/when the next handler
+// fails to handle the message.  The parameter source is the source to use in
+// the response message.
+func New(next, egress wrpkit.Handler, source string, logger *zap.Logger, opts ...Option) (*Handler, error) {
+	// TODO - load config from file system
+
+	h := Handler{
+		next:   next,
+		egress: egress,
+		source: source,
```
denopink, line 84: looks like you forgot to add the logger

I think using `logger := logger.Named("mocktr181")` would be helpful during debugging

denopink: i think adding the logger as an option is better

piccione99: oops, thanks, good catch

cmd/xmidt-agent/config.go
```diff
@@ -173,6 +174,10 @@ type Storage struct {
 	Durable   string
 }
 
+type MockTr181 struct {
+	FilePath string
```
schmidtw, line 178: I think we should have an `Enabled` field to turn this functionality on or off explicitly.

piccione99: oooh, I thought there was going to be some config for each environment / scenario that specified which wrphandlers we wanted.  Similar to java spring (sorry if that is an insulting comparison)

internal/wrphandlers/mocktr181/handler.go
```diff
@@ -0,0 +1,211 @@
...
+type optionFunc func(*Handler) error
+
+func (f optionFunc) apply(c *Handler) error {
+	return f(c)
+}
+
+// Handler sends a response when a message is required to have a response, but
+// was not handled by the next handler in the chain.
+type Handler struct {
+	next       wrpkit.Handler
```
schmidtw: Since this is a "final endpoint" that doesn't forward on to any other links in the chain - I don't think it needs a "next", right?

cmd/xmidt-agent/config.go
schmidtw: Basically yes, like Java spring.  An `Enabled` field will make it easy to configure the module on or off because the structure will have default values & be present in all configuration.

schmidtw (approved):

PR #60, opened 2024-04-04
Change the flags that enable/disable a component use 'Enabled' for co…

schmidtw:
…nsistency.

### DIFF
--- a/cmd/xmidt-agent/config.go
+++ b/cmd/xmidt-agent/config.go
@@ -31,8 +31,8 @@ type Config struct {
 }
 
 type Websocket struct {
-	// Disable determines whether or not to disable xmidt-agent's websocket
-	Disable bool
+	// Enabled determines whether or not to enable xmidt-agent's websocket
+	Enabled bool
 	// URLPath is the device registration url path
 	URLPath string
 	// AdditionalHeaders are any additional headers for the WS connection.
--- a/cmd/xmidt-agent/config.go
+++ b/cmd/xmidt-agent/config.go
@@ -258,6 +258,7 @@ var defaultConfig = Config{
 		},
 	},
 	Websocket: Websocket{
+		Enabled:               true,
 		URLPath:               "api/v2/device",
 		FetchURLTimeout:       30 * time.Second,
 		PingInterval:          30 * time.Second,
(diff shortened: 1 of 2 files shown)


denopink (approved):
I chose `Disable` because it worked nicely with the default value, but understand and agree with the need to keep things consistent 

lgtm

cmd/xmidt-agent/config.go
```diff
@@ -258,6 +258,7 @@ var defaultConfig = Config{
 		},
 	},
 	Websocket: Websocket{
+		Enabled:               true,
 		URLPath:               "api/v2/device",
```
denopink, line 262: added a new default if we wanted to ws to be enable by default
https://github.com/xmidt-org/xmidt-agent/pull/60#pullrequestreview-1980095677

piccione99: this is fine but does disabled make more sense because default bool is false? I missed this pattern so I can switch the mocktr181 to use disabled if that's better

denopink: fyi @schmidtw in case you didn't want this behavior

denopink: because of the power of goschtalt, either one is fine and both works for me

schmidtw: Hmm - the question I have is: 
If this defaults to on, will the default configuration actually work for any service?

If you need to provide some configuration to make ti work, I think defaulting it to off is ok.  If it will work just fine without any non-default values, then I agree that `Disabled` would make more sense.

denopink: ~~what do you think @johnabass ?~~ nevermind

schmidtw: @denopink and I chatted & while well intended, I'm not merging this PR for a few reasons:

1. When we enable QUIC as a protocol, we'll know better how we want to refactor the config & yes, it will be a breaking change, but that's probably better than guessing about several things and being wrong & then still have a breaking change later.
2. This doesn't help us reach our initial goal/deadline.

schmidtw: Thanks for the feedback!

PR #63, opened 2024-04-04
feat: integrate wrphandlers

denopink:


cmd/xmidt-agent/config.go
```diff
@@ -296,6 +302,9 @@ var defaultConfig = Config{
 			MaxInterval: 341*time.Second + 333*time.Millisecond,
 		},
 	},
+	Pubsub: Pubsub{
+		PublishTimeout: 200 * time.Millisecond,
+	},
 	MockTr181: MockTr181{
 		FilePath: "./mock_tr181.json",
 		Enabled:  true,
```
denopink: I'm assuming we don't want MockTr181 to run by default
```suggestion
		Enabled:  false,
```

internal/websocket/e2e_test.go
```diff
@@ -256,7 +266,7 @@ func TestEndToEndConnectionIssues(t *testing.T) {
 				require.NoError(err)
 				defer c.CloseNow()
 
-				ctx, cancel := context.WithTimeout(r.Context(), 2000000*time.Millisecond)
+				ctx, cancel := context.WithTimeout(r.Context(), 200*time.Millisecond)
```
denopink, line 269: fixed a silly debugging artifact from my last pr

internal/websocket/ws.go
```diff
@@ -159,6 +158,10 @@ func (ws *Websocket) Start() {
 // Stop stops the websocket connection.
 func (ws *Websocket) Stop() {
 	ws.m.Lock()
+	if ws.conn != nil {
+		ws.conn.Close(nhws.StatusNormalClosure, "")
+	}
```
denopink, line 165: at the very least it helps with testing

internal/websocket/ws.go
```diff
@@ -169,6 +172,12 @@ func (ws *Websocket) Stop() {
 	ws.wg.Wait()
 }
 
+// AddMessageListener adds a message listener to the WS connection.
+// The listener will be called for every message received from the WS.
+func (ws *Websocket) AddMessageListener(listener event.MsgListener, cancel ...*event.CancelFunc) event.CancelFunc {
+	return event.CancelFunc(ws.msgListeners.Add(listener))
+}
+
```
denopink, line 186: We needed a way to add listeners after the fact, i.e. adding a ws - to - wrphandler adapter

cmd/xmidt-agent/wrphandlers.go
```diff
@@ -0,0 +1,208 @@
...
+		pubsub.WithPublishTimeout(in.Pubsub.PublishTimeout),
+		pubsub.WithEgressHandler(in.WSHandler, &egress),
+	}
+
+	var ps *pubsub.PubSub
+	if in.MockTr181.Enabled {
+		var mocktr pubsub.CancelFunc
+
+		mockDefaults := []mocktr181.Option{
+			mocktr181.FilePath("mock_tr181_test.json"),
```
piccione99: can we use mock_tr181.json here?  The the test one is for unit tests.

piccione99: Also, the default for mocktr181 is specified in config.go, should that be used here?

denopink: good catch, totally forgot to remove this

cmd/xmidt-agent/wrphandlers.go
```diff
@@ -0,0 +1,208 @@
...
+	return wshandler.New(in.WS)
+}
+
+type missingIn struct {
+	fx.In
+
+	// Configuration
+	// Note, DeviceID and PartnerID is pulled from the Identity configuration
+	DeviceID wrp.DeviceID
+	WSConfg  Websocket
```
schmidtw: I don't think the WS should be needed.  Missing should depend on auth.

cmd/xmidt-agent/wrphandlers.go
```diff
@@ -0,0 +1,208 @@
...
+
+	// wrphandlers
+	WSHandler *wshandler.Handler
+	Pubsub    *pubsub.PubSub
+}
+
+func provideMissingHandler(in missingIn) (h *missing.Handler, err error) {
+	defer func() {
+		if err != nil {
```
schmidtw: Instead of doing this as a defer, why not make the conditional after the code it handles?  This brings more mental baggage for me.

schmidtw (approved):

PR #65, opened 2024-04-05
Fork the upstream websocket code.

schmidtw:
We need to fork the WS code because we need to support ping and pong functionality and the upstream code can't support that yet.


schmidtw:
This is not perfect, but it provides a pretty simple way to achieve what we need to ship.

PR #68, opened 2024-04-05
Add more sane default values, and move the logging configuration into…

### LINKED ISSUE #67: Figure out how to configure the log rotation.

schmidtw:
… the default instead of built in.  This should fix #67.

### DIFF
--- a/cmd/xmidt-agent/config.go
+++ b/cmd/xmidt-agent/config.go
@@ -340,9 +340,9 @@ var defaultConfig = Config{
 			EncodeCaller:   "short",
 		},
 		Rotation: &sallust.Rotation{
-			MaxSize:    1  // 1MB max/file
-			MaxAge:     30,              // 30 days max
-			MaxBackups: 10,              // max 10 files
+			MaxSize:    1,  // 1MB max/file
+			MaxAge:     30, // 30 days max
+			MaxBackups: 10, // max 10 files
 		},
 	},
 	MockTr181: MockTr181{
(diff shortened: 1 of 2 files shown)


denopink (approved):
lgtm!

cmd/xmidt-agent/config.go
```diff
@@ -307,6 +324,27 @@ var defaultConfig = Config{
...
+			MessageKey:     "M",
+			StacktraceKey:  "S",
+			LineEnding:     zapcore.DefaultLineEnding,
+			EncodeLevel:    "capital",
+			EncodeTime:     "RFC3339Nano",
+			EncodeDuration: "string",
+			EncodeCaller:   "short",
+		},
+		Rotation: &sallust.Rotation{
+			MaxSize:    1 * 1024 * 1024, // 1MB max/file
```
denopink:
```suggestion
			MaxSize:    1  // 1MB max/file
```

looks like MaxSize's unites are already in megabytes https://pkg.go.dev/gopkg.in/natefinch/lumberjack.v2#section-readme
```
        // MaxSize is the maximum size in megabytes of the log file before it gets
	// rotated. It defaults to 100 megabytes.
```

denopink (approved):
just 1 suggested fix, otherwise lgtm! 🍻

denopink:
looks like this pr closes #67

cmd/xmidt-agent/config.go
schmidtw: Good catch, thank you.

PR #89, opened 2024-04-12
Interface

piccione99:
Added a service to make network calls. We can always swap out the actual network library if we don't what this one returns.   This one works cross platform but I can also pull in the linux only version.


internal/net/network_service.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+var NetworkServiceModule = fx.Module("networkService",
+	fx.Provide(
+		func(in NetworkServiceIn) *NetworkService {
+			return &NetworkService{
+				n: NewNetworkWrapper(),
+			}
+		}),
+)
```
denopink: this repo started keeping the `fx` related code in the `cmd/xmidt` as their own files.

I don't have a preference between the current or this pr's pattern, but I'm assuming we want to keep things consistent.

denopink (approved):
left 1 style suggestion, otherwise lgtm

internal/net/network_service.go
piccione99: good point. I'll switch it.  we are def not consistent between repos

schmidtw (approved):

schmidtw (approved):

PR #95, opened 2024-04-17
feat: qos integration

denopink:


cmd/xmidt-agent/ws.go
```diff
@@ -32,15 +49,18 @@ type wsIn struct {
...
+
+	// eventor adapters
+	WRPHandlerAdapterCancel event.CancelFunc `name:"wrphandlerAdapter"`
+	QOSStartAdapter         event.CancelFunc `name:"qosStartAdapter"`
+	QOSEndAdapter           event.CancelFunc `name:"qosEndAdapter"`
 }
 
 type wsOut struct {
 	fx.Out
-	WSHandler               wrpkit.Handler
```
denopink, line 39:
```
	WSHandler               wrpkit.Handler
```
should have been removed in my last pr

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	"github.com/xmidt-org/wrp-go/v3"
+)
+
+// Item contains a wrp Message pointer and is managed by PriorityQueue.
+type Item struct {
+	value    *wrp.Message
+	priority wrp.QOSValue
+	index    int
```
schmidtw: Why is there an index?  This seems like it's not really important.

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// PriorityQueue implements heap.Interface and holds Items, using wrp.QOSValue as its priority.
+// https://xmidt.io/docs/wrp/basics/#qos-description-qos
+type PriorityQueue []*Item
```
schmidtw: Why are the Items pointers vs values?  Values are easier to work with.

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -0,0 +1,60 @@
...
+package qos
+
+import (
+	"github.com/xmidt-org/wrp-go/v3"
+)
+
+// Item contains a wrp Message pointer and is managed by PriorityQueue.
+type Item struct {
+	value    *wrp.Message
+	priority wrp.QOSValue
```
schmidtw: I wouldn't duplicate this, but use the value in the value when needed.

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -0,0 +1,60 @@
...
+
+	*pq = append(*pq, i)
+}
+
+func (pq *PriorityQueue) Pop() any {
+	if len(*pq) == 0 {
+		return nil
+	}
+
+	old := *pq
```
schmidtw: basically all of this can be deleted if we use slices

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,206 @@
...
+	case <-h.empty:
+	}
+
+	h.items <- append(items, true)
+}
+
+// trimQueue pops the message with lowest priority and last in line
+func (h *Handler) trimQueue() {
+	// when maxQueueSize is 0, len(h.pq) > 1 allows qos to send 1 message at a time
+	for len(h.pq) > 1 && (int(unsafe.Sizeof(h.pq)) > h.maxQueueSize) {
```
denopink: @schmidtw or @johnabass do you know of quick way to get the size of an array and the data the pointers are pointing to?... i can't remember for the life of me

internal/wrphandlers/qos/priority_queue.go
schmidtw, line 1: I think the pq needs refining to make it really useful.  I would put the logic of size management and prioritization into the queue itself & pull it out of the other structs.

internal/wrphandlers/qos/qos.go
schmidtw, line 1: I'm going to stop here & see what we want to do with the pq.

internal/wrphandlers/qos/priority_queue.go
denopink: forgot to remove this, used it for a quick test

internal/wrphandlers/qos/priority_queue.go
denopink: for sure, we can use values

internal/wrphandlers/qos/qos.go
schmidtw: Not that I'm aware of.  I would create a function that takes a wrp.Message and calculates the approximate size based on the sizes of the strings and byte arrays.  I don't think being super accurate is as important as "generally adding things up" since go may be reusing pointers under the covers, etc.

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,192 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package qos
+
+import (
+	"container/heap"
+	"context"
+	"errors"
+	"sync"
+	"unsafe"
```
schmidtw: I don't think we should have any `unsafe` code in this component.

internal/wrphandlers/qos/qos.go
johnabass: The way to do this is with `unsafe`.  I know the name might be scary, but it's the package that has memory calculations in it.

But the bigger issue is:  Why calculate the message size?  Why not simply base it on the size of the payload?  The struct fields themselves don't contribute much to the overall message size.

cmd/xmidt-agent/config.go
```diff
@@ -366,4 +378,10 @@ var defaultConfig = Config{
 		FilePath:    "./mock_tr181.json",
 		ServiceName: "config",
 	},
+	QOS: QOS{
+		Queue: QOSQueue{
+			MaxQueueSize:  1 * 1024 * 1024, // 1MB max/queue,
+			MaxQueueDepth: 1000,            // 1000 messages
```
denopink, line 384: @schmidtw in terms of rdk, what are good defaults for these?

schmidtw: those are fine

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -0,0 +1,95 @@
...
+func (pq *PriorityQueue) Enqueue(msg wrp.Message) {
+	pq.m.Lock()
+	heap.Push(pq, msg)
+	pq.trim()
+	pq.m.Unlock()
+}
+
+// trim removes the lowest priority message if the queue is full.
+func (pq *PriorityQueue) trim() {
+	// when pq.IsFull() is true, pq.Len() > 1 ensures at least 1 message is queued
```
schmidtw: This is fine, but it has the side effect of 1 message being present can be of any size.

cmd/xmidt-agent/config.go
```diff
@@ -37,9 +37,21 @@ type Config struct {
...
 
+type QOS struct {
+	Queue QOSQueue
+}
+
+type QOSQueue struct {
+	// MaxQueueSize is the allowable max size of the queue based on the sum of all queued wrp message's payloads
+	MaxQueueSize int
+	// MaxQueueDepth is the total wrp message allowed
+	MaxQueueDepth int
```
schmidtw, line 52: The depth causes some problems later - I'd focus on the size and drop the depth.

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -0,0 +1,95 @@
...
+	msg, ok := heap.Pop(pq).(wrp.Message)
+
+	return msg, ok
+
+}
+
+// Enqueue queues the given message.
+func (pq *PriorityQueue) Enqueue(msg wrp.Message) {
+	pq.m.Lock()
+	heap.Push(pq, msg)
```
schmidtw: This creates an unexpected problem - this will increase the capacity even when you don't want to.

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -0,0 +1,95 @@
...
+	var s int
+	for _, msg := range pq.queue {
+		s += len(msg.Payload)
+	}
+
+	return s
+}
+
+func (pq *PriorityQueue) IsFull() bool {
+	return len(pq.queue) == cap(pq.queue) || pq.Size() > pq.maxQueueSize
```
schmidtw, line 77: `cap()` can be much larger than the MaxQueueDepth value because each time the size needs to be increased when `append()` is called, the `cap()` will increase, too.  Best to just drop this field entirely.

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,161 @@
...
+				return nil, err
+			}
+		}
+	}
+
+	return h, nil
+}
+
+// Start starts the qos queue ingestion.
+func (h *Handler) Start() {
```
schmidtw, line 71: `Start()` is no longer needed.  Since messages are drained upon insertion onto the queue or when the WS is available, this can be entirely removed.

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+// Handler queues incoming messages and sends them to the next wrphandler
+type Handler struct {
+	next wrpkit.Handler
+	// queue for wrp messages
+	queue PriorityQueue
+
+	m sync.Mutex
+	// ingestingQueue blocks if the SendQueuedMessages() is already running, i.e.: send
```
schmidtw, line 39: None of these (`ingestingQueue`, `shutdown`, `ctx`) fields are needed because the `queue` being empty governs everything.  Since the queue is added to or removed from, then unlocked in between, it gives the cleanup process the ability to grab the mutex & dump everything right then.  Everything stops because there is no work.

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,161 @@
...
+	h.ctx, h.shutdown = context.WithCancel(context.Background())
+	h.m.Unlock()
+
+	// send as many queued messages as possible
+	go h.SendQueuedMessages()
+
+}
+
+// Stop stops the qos queue ingestion.
+func (h *Handler) Stop() {
```
schmidtw, line 87: `Stop()` should change to something like `Cleanup()` or `Empty()` or `Drain()` because that's all it's going to do ... just throw away everything in the PQ.

Instead of calling `shutdown()`, just rapid fire `Pop()` everything off the heap, or replace the heap with an empty one.  Do this in the mutex lock region.

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,161 @@
...
+	return nil
+}
+
+// SendQueuedMessages sends as many queued messages as possible until the following occurs:
+// 1. the qos context was cancelled
+// 2. qos queue is empty
+// 3. there was a delivery failure
+// Note, Start() will automatically trigger a SendQueuedMessages().
+func (h *Handler) SendQueuedMessages() {
+	defer func() {
```
schmidtw, line 116: No defer should be needed after the `ingestingQueue`, `ctx`, etc are removed.

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,161 @@
...
+	default:
+		// a ingestingQueue is blocked, a SendQueuedMessages() already in progress
+		return
+	}
+
+	h.m.Lock()
+	ctx := h.ctx
+	h.m.Unlock()
+
+	for {
```
schmidtw, line 138: I'd change the naked `for` to be conditional based on if there is a message:

```go
for msg, ok := h.queue.Dequeue(); ok; {
...
}
```

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,161 @@
...
+		// always get the next highest priority message
+		msg, ok := h.queue.Dequeue()
+		if !ok {
+			// queue is empty
+			return
+		}
+
+		select {
```
schmidtw, line 146: There's no need for this since an empty queue defines "we're done".

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -0,0 +1,95 @@
...
+
+// Dequeue returns the next highest priority message.
+func (pq *PriorityQueue) Dequeue() (wrp.Message, bool) {
+	defer pq.m.Unlock()
+	pq.m.Lock()
+	if pq.Len() == 0 {
+		return wrp.Message{}, false
+	}
+
+	msg, ok := heap.Pop(pq).(wrp.Message)
```
schmidtw, line 32: I'd not assign these, but just return from the `Pop()`

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -0,0 +1,95 @@
...
+	pq.queue[i], pq.queue[j] = pq.queue[j], pq.queue[i]
+}
+
+func (pq *PriorityQueue) Push(x any) {
+	pq.queue = append(pq.queue, x.(wrp.Message))
+}
+
+func (pq *PriorityQueue) Pop() any {
+	old := pq.queue
+	n := len(old)
```
schmidtw: I'd refactor this code like this:

```go
last := len(pq.queue) - 1
if last < 0 {
    return nil
}

item := pq.queue[last]
// avoid memory leak
pq.queue[last] = wrp.Message{}

pq.queue = pq.queue[0:last]

return item
```

There isn't really an old structure, you've just got 2 names for it.  `n` is really the last item in the list's index, so `n-1` everywhere is a bit tedious.

schmidtw (changes requested):
Most of these comments center around there being a simpler way to do the same thing.  Less code makes for less tests, etc.

internal/wrphandlers/qos/priority_queue.go
denopink: oh man, good catch

internal/wrphandlers/qos/qos.go
denopink: got it. I was considering a situation where we have queued messages and the WS closes and reconnects, leaving the messages queued until the agent receives a new message... how long could that be (minutes or hours?) and do we care? or maybe we want a linger for the queue if the ws is connected but we haven't received a message for awhile?

internal/wrphandlers/qos/qos.go
denopink: I think I understand your suggestion, but I have a few clarifying questions

> None of these (ingestingQueue, shutdown, ctx) fields are needed

in terms of `ingestingQueue`, I was trying to limit how many parallel `go h.SendQueuedMessages()`  goroutines can be ran, since our websocket locks during writes. From my understanding that meant we can potentially have the following problems if don't use `ingestingQueue`:
- unbounded growth of `go h.SendQueuedMessages()`  goroutines (since we call it every time we receive a message from xmidt)
- a group of goroutines waiting to write to the ws one at time, defeating the purpose of having more than 1 goroutine
 

>Since the queue is added to or removed from, then unlocked in between, it gives the cleanup process the ability to grab the mutex & dump everything right then. 

I'm assuming this is related to this comment as well https://github.com/xmidt-org/xmidt-agent/pull/95/files#r1571847953:

> 
> Stop() should change to something like Cleanup() or Empty() or Drain() because that's all it's going to do ... just throw away everything in the PQ.
> 
> Instead of calling shutdown(), just rapid fire Pop() everything off the heap, or replace the heap with an empty one. Do this in the mutex lock region.
> Comment
> 

what would trigger the cleanup process then? 

I thought the cleanup process, dumping everything without attempting to send the messages, would naturally be the go program or the fx app restarting. 

right now `Stop` is called when a disconnect event occurs in order to stop a running `go h.SendQueuedMessages()` goroutine with `shutdown, ctx` (if there is 1 running), but we keep the queue as is and wait until the next reconnect event occur in order to start ingesting the queue again (starting where we left off). 

I assumed that's what we wanted (to hold onto messages between disconnects) since that's the only reason (from my understanding) why messages would fail to send

That's why I used `shutdown, ctx` as the stopping mechanism for the queue rather them dumping everything

internal/wrphandlers/qos/priority_queue.go
denopink: yeah I wanted to do that but go doesn't let us do that and keep the boolean indicator of whether the assertion succeeded

image

schmidtw: Interesting

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -0,0 +1,106 @@
...
+
+	msg, ok := heap.Pop(pq).(wrp.Message)
+
+	return msg, ok
+
+}
+
+// Enqueue queues the given message.
+func (pq *PriorityQueue) Enqueue(msg wrp.Message) {
+	defer pq.m.Unlock()
```
denopink: @schmidtw here's the update to how we trim the queue, tested with cd and it looks like it's working as expected now

PR #104, opened 2024-04-19
feat: qos integration

### LINKED ISSUE #98: feat: qos integration

denopink:
100% test coverage


internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,153 @@
...
+// Stop stops the qos queue ingestion.
+func (h *Handler) Stop() {
+	h.m.Lock()
+	shutdown := h.shutdown
+	// allows qos to restart
+	h.shutdown = nil
+	if shutdown == nil {
+		return
+	}
+	h.m.Unlock()
```
schmidtw: This creates a deadlock because you returned before the lock was released.

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,153 @@
...
+		case <-h.ingestQueue:
+		// queue ingestion will start
+		case <-ctx.Done():
+			// QOS was stopped, exit
+			return
+		}
+
+		// always get the next highest priority message
+		// `ok` will stop the loop if the queue is empty
+		for msg, ok := h.queue.Dequeue(); ok; msg, ok = h.queue.Dequeue() {
```
schmidtw: It looks like you're Dequeuing 2 messages & throwing one away.

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,153 @@
...
+			select {
+			case <-ctx.Done():
+				// QOS was stopped, re-enqueue message and exit
+				h.queue.Enqueue(msg)
+				return
```
schmidtw, line 162: You don't want to return here because you exit the run loop which won't get restarted.

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,153 @@
...
+
+// run sends as many queued messages as possible until the following occurs:
+// 1. the qos context was cancelled
+// 2. qos queue is empty
+// 3. there was a delivery failure
+// Note, Start() will automatically trigger a SendQueuedMessages().
+func (h *Handler) run(ctx context.Context) {
+	for {
+		select {
+		case <-h.ingestQueue:
```
schmidtw: You're missing the trigger based on the WS connected listener.

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,153 @@
...
+		// queue ingestion will start
+		case <-ctx.Done():
+			// QOS was stopped, exit
+			return
+		}
+
+		// always get the next highest priority message
+		// `ok` will stop the loop if the queue is empty
+		for msg, ok := h.queue.Dequeue(); ok; msg, ok = h.queue.Dequeue() {
+			select {
```
schmidtw: I'd change the select to the following because it's simpler & doesn't have the bug the return creates of ending the go routune

```go
if ctx.Err() != nil {
  h.queue.Enqueue(msg)
  break
}
```

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,153 @@
...
+	case h.ingestQueue <- true:
+		// send as many queued messages as possible
+	default:
+		// queue ingestion is already in progress
+	}
+
+}
+
+// Stop stops the qos queue ingestion.
+func (h *Handler) Stop() {
```
schmidtw, line 89: It's important to note that `Stop()` makes no assurances that the goroutine is actually shut down before the method returns.  This isn't technically a problem, but it can cause some interesting scenarios when you go to test it.

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	return h, nil
+}
+
+// Start starts the qos queue ingestion.
+func (h *Handler) Start() {
+	h.m.Lock()
+	if h.shutdown != nil {
+		return
```
schmidtw: This will deadlock the code because you exited the code block without releasing the mutex.

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -0,0 +1,95 @@
...
+	queue []wrp.Message
+	// maxQueueSize is the allowable max size of the queue based on the sum of all queued wrp message's payloads
+	maxQueueSize int
+
+	m sync.Mutex
+}
+
+// Dequeue returns the next highest priority message.
+func (pq *PriorityQueue) Dequeue() (wrp.Message, bool) {
+	defer pq.m.Unlock()
```
schmidtw: Usually this is done in the other order in go.  (`Lock()`, then defer `Unlock()`)

schmidtw (changes requested):

internal/wrphandlers/qos/qos.go
denopink: I think that's covered by the connect listener, we restart the qos and it's queue ingestion
```
in.WS.AddConnectListener(
		event.ConnectListenerFunc(func(event.Connect) {
			h.Start()
		}),
		&in.QOSStartAdapter,
	)
```

internal/wrphandlers/qos/qos.go
denopink: I think that's covered by the connect listener, we restart the qos and it's queue ingestion

internal/wrphandlers/qos/qos.go
denopink: that's such a great point... interesting

at least `case h.ingestQueue <- true` will be blocked until the old goroutine is shutdown via the ctx cancel (the new goroutine would be created on connect event)

i can see this becoming dangerous in situtations where the goroutine code has a bug and doesn't shutdown.

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,161 @@
...
+
+	return nil
+}
+
+// run sends as many queued messages as possible until the following occurs:
+// 1. the qos context was cancelled
+// 2. qos queue is empty
+// 3. there was a delivery failure
+// Note, Start() will automatically trigger a SendQueuedMessages().
+func (h *Handler) run(ctx context.Context) {
```
johnabass: You can simplify this code dramatically by just spawning (1) longrunning goroutine.  The complexity of managing the goroutine's lifecycle is definitely codesmell that indicates the code is working too hard.

You basically have (3) signals to manage:
- a done channel, so that you can control the goroutine's shutdown
- a message channel
- a "connection ready" channel for the websocket

You'll need a `Connection` type with a method that returns a signaling channel that indicates the IO is done.  `gorilla/mux` and `gobwas` have this, but not sure if the ws library you're using has it.  You can roll your own with something like:

```
type Connection struct {
    l sync.Mutex
    w io.Writer
}

func (c *Connection) Write(m wrp.Message) <-chan struct{} {
    // will probably want to protect this with a counting semaphore
    // and/or just have a longrunning goroutine that services the connection
    ready := make(chan struct{})
    go func() {
        defer close(ready)

        c.l.Lock()
        defer c.l.Unlock()
        c.w.Write(/*whatever*/)
    }()

    return ready
}
```

Pseudocode:

```
func serviceQOS(done <-chan struct{}, msg <- chan wrp.Message, conn *Connection) {
    // create and manage the priority queue here:
    var pq PriorityQueue

    // signaling channel from the connection
    var ready <-chan struct{}

    for {
        select {
            case <- done:
                return

            case m := <-msg:
                pq.Push(m)
                if ready == nil {
                     top, _ := pq.Pop()
                     ready = conn.Write(top)
                }

             case <-ready:
                 ready = nil
                 if top, ok := pq.Pop(); ok {
                     ready := conn.Write(top)
                 }
        }
    }
}
```

With the above, you're basically leveraging the go runtime to do most of the things this PR is doing.

cmd/xmidt-agent/config.go
```diff
@@ -37,9 +37,21 @@ type Config struct {
...
 
+type QOS struct {
+	Queue QOSQueue
+}
+
+type QOSQueue struct {
+	// MaxQueueSize is the allowable max size of the queue based on the sum of all queued wrp message's payloads
+	MaxQueueSize int
+	// MaxQueueDepth is the total wrp message allowed
+	MaxQueueDepth int
```
denopink: removing MaxQueueDepth

cmd/xmidt-agent/main.go
```diff
@@ -224,7 +226,7 @@ func onStart(cred *credentials.Credentials, ws *websocket.Websocket, logger *zap
 	}
 }
 
-func onStop(ws *websocket.Websocket, shutdowner fx.Shutdowner, eventCancelList []event.CancelFunc, pubsubCancelList []pubsub.CancelFunc, logger *zap.Logger) func(context.Context) error {
+func onStop(ws *websocket.Websocket, shutdowner fx.Shutdowner, eventCancelList []event.CancelFunc, pubsubCancelList []pubsub.CancelFunc, cancels []func(), logger *zap.Logger) func(context.Context) error {
```
denopink: in a separate pr, I want to simplify this to `onStop(ws *websocket.Websocket, shutdowner fx.Shutdowner, cancels []func(), logger *zap.Logger) func(context.Context)`, replacing `eventCancelList` and `pubsubCancelList` with `cancels`.

https://github.com/xmidt-org/xmidt-agent/issues/109

cmd/xmidt-agent/ws.go
```diff
@@ -36,7 +35,6 @@ type wsIn struct {
 
 type wsOut struct {
 	fx.Out
-	WSHandler               wrpkit.Handler
```
denopink, line 39: unused

internal/wrphandlers/qos/qos_test.go
```diff
@@ -0,0 +1,4 @@
+// SPDX-FileCopyrightText: 2024 Comcast Cable Communications Management, LLC
+// SPDX-License-Identifier: Apache-2.0
+
+package qos_test
```
denopink, line 4: i connected to cd xmidt cluster, and the agent works and the heap trimming also works as expected

I'm adding tests in the next batch of commits

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,134 @@
...
+// Note, New will automatically start the serviceQOS goroutine.
+func serviceQOS(done <-chan struct{}, queue <-chan wrp.Message, maxHeapSize int, handleWRP serviceQOSHandler) {
+	// create and manage the priority queue
+	pq := PriorityQueue{maxHeapSize: maxHeapSize}
+	var (
+		// signaling channel from the handleWRP
+		ready <-chan struct{}
+		// channel for failed deliveries, re-enqueue message
+		failedMsg <-chan *wrp.Message
+	)
```
denopink, line 108: @johnabass i was thinking we want to keep those two functionality separate, a signaling channel and a channel for failed deliveries (in order to re-enqueue the failed message), that way it's easier to read and reason

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,138 @@
...
+// serviceQOS is stopped when Handler.Cancel() is called, closing the `done` chan.
+// Note, New will automatically start the serviceQOS goroutine.
+func serviceQOS(done <-chan struct{}, queue <-chan wrp.Message, maxHeapSize int, handleWRP serviceQOSHandler) {
+	// create and manage the priority queue
+	pq := PriorityQueue{maxHeapSize: maxHeapSize}
+	var (
+		// Signaling channel from the handleWRP.
+		ready <-chan struct{}
+		// Channel for failed deliveries, re-enqueue message.
+		failedMsg <-chan wrp.Message
```
denopink, line 151: @johnabass i was thinking we want to keep those two functionality separate, a signaling channel and a channel for failed deliveries (in order to re-enqueue the failed message), that way it's easier to read and reason

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// HandleWRP queues incoming messages while the background serviceQOS goroutine attempts
+// to send as many queued messages as possible, where the highest QOS messages are prioritized
+func (h *Handler) HandleWrp(msg wrp.Message) error {
+	// h.queue will never block as long as the serviceQOS goroutine is running.
+	h.queue <- msg
+
+	return nil
+}
```
denopink: @johnabass  should we check whether `serviceQOS`'s `done <-chan struct{}` has been closed? that way `handler.HandleWrp` can return an `io.EOF` like error when someone tries to write a message when qos's ingestion has stop.
I was thinking that'll be better than just halting because the agent's qos calls are not asynchronous, so it blocks the agent's websocket read.

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,138 @@
...
+		// Channel for failed deliveries, re-enqueue message.
+		failedMsg <-chan wrp.Message
+	)
+
+	for {
+		select {
+		case <-done:
+			return
+		case msg := <-queue:
+			pq.Enqueue(msg)
```
denopink: @johnabass is it ever a good idea to add additional safe guards like this, even tho we aren't closing `queue <-chan wrp.Message` at any point?

```suggestion
		case msg, ok := <-queue:
			if !ok {
				return
			}
			
			pq.Enqueue(msg)
```

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -0,0 +1,103 @@
...
+
+	msg, ok := heap.Pop(pq).(wrp.Message)
+
+	return msg, ok
+}
+
+// Enqueue queues the given message.
+func (pq *PriorityQueue) Enqueue(msg wrp.Message) {
+	// Check whether msg would already violate maxHeapSize.
+	if len(msg.Payload) >= pq.maxHeapSize {
```
```suggestion
	if len(msg.Payload) > pq.maxHeapSize {
```

denopink: @johnabass besides checking  whether msg would already violate `maxHeapSize`, should we consider enforcing a max size for a given message?

Otherwise, a non-empty queue may be emptied out in order to queue an incoming message (with a similar or great `QOS` value) of size `maxHeapSize`

internal/wrphandlers/qos/qos.go
denopink: although i can see having multiple `return`(in this context) becoming a little confusing for anyone not familiar with the qos code

internal/wrphandlers/qos/options.go
```diff
@@ -0,0 +1,22 @@
...
+	"fmt"
+)
+
+// MaxHeapSize is the allowable max size of the qos' priority queue, based on the sum of all queued wrp message's payload
+func MaxHeapSize(s int) Option {
+	return optionFunc(
+		func(h *Handler) error {
+			if s < 0 {
+				return fmt.Errorf("%w: negative MaxHeapSize", ErrMisconfiguredQOS)
+			}
```
denopink: @johnabass should we allow the zero value? If so, I'm assuming we'll treat zero as a PriorityQueue with unlimited heap size.

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -0,0 +1,103 @@
...
+		return wrp.Message{}, false
+	}
+
+	msg, ok := heap.Pop(pq).(wrp.Message)
+
+	return msg, ok
+}
+
+// Enqueue queues the given message.
+func (pq *PriorityQueue) Enqueue(msg wrp.Message) {
```
denopink: I haven't yet added the zero behavior for `maxQueueSize`
https://github.com/xmidt-org/xmidt-agent/pull/104/files#r1578469699

internal/wrphandlers/qos/qos.go
johnabass: You can't check if a channel is closed.  A channel becomes signalled when it's closed, which means a `case <-done` will trigger.

If you're wanting to gate the handler, so that it returns an an error when the goroutine isn't running, a good old-fashioned mutex works.  You probably should have that anyway, to make start/stop atomic:

```
func (h *Handler) HandleWrp(msg wrp.Message) (err error) {
        h.state.Lock()
        defer h.state.Unlock()
        
	if h.queue != nil {
              // when stopped, set h.queue = nil under the same lock
	      h.queue <- msg
        } else {
              err = SomeAppropriateError
        }

	return
}
```

denopink: I actually removed the `start` and `stop` functions to simplify qos

totally that makes sense, I was originally thinking of storing `ctx.Done()` a part of Handler to determine if qos has been shutdown.

In a situation like this, where `ctx.Done()` is basically telling us this qos is dead... create a new qos, would we want to re-use the same `ctx.Done()` that `serviceQOS` is using? I was think that would be cleaner.

```
func (h *Handler) HandleWrp(msg wrp.Message) error {
	switch {
	case <-h.done:
		return SomeAppropriateError
	default:
		// h.queue will never block as long as the serviceQOS goroutine is running.
		h.queue <- msg
	}

	return nil
}
```

denopink: @johnabass 

I actually removed the `start` and `stop` functions to simplify qos

totally that makes sense, I was originally thinking of storing `ctx.Done()` a part of Handler to determine if qos has been shutdown.

In a situation like this, where `ctx.Done()` is basically telling us this qos is dead... unless a new qos created, would we want to re-use the same `ctx.Done()` that `serviceQOS` is using? I was think that would be cleaner.

```
func (h *Handler) HandleWrp(msg wrp.Message) error {
	select {
        // h.done is the same `done <-chan struct{}` that serviceQOS is using
	case <-h.done:
		return SomeAppropriateError
	default:
		// h.queue will never block as long as the serviceQOS goroutine is running.
		h.queue <- msg
	}

	return nil
}
```

internal/wrphandlers/qos/qos_test.go
denopink: adding `serviceQOS` tests next

internal/wrphandlers/qos/priority_queue_test.go
```diff
@@ -0,0 +1,4 @@
+// SPDX-FileCopyrightText: 2024 Comcast Cable Communications Management, LLC
+// SPDX-License-Identifier: Apache-2.0
+
+package qos_test
```
denopink: wip

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -0,0 +1,103 @@
...
+	var s int
+	for _, msg := range pq.heap {
+		s += len(msg.Payload)
+	}
+
+	return s
+}
+
+func (pq *PriorityQueue) IsFull() bool {
+	return len(pq.heap) == cap(pq.heap) || pq.Size() > pq.maxQueueSize
```
denopink: forgot to update this
```suggestion
	return pq.Size() > pq.maxQueueSize
```

internal/wrphandlers/qos/qos.go
johnabass: You can't reuse a context or it's done channel.  You have to create a new context and use that.  That's one reason why you need some atomic state management around starting/stopping the qos goroutine.

internal/wrphandlers/qos/options.go
johnabass: Yes, (almost) always use your zero value for something useful, like a default.  I would never have an unbounded queue though, especially on a CPE.  I would just have a DefaultQueueSize that got used when s==0.

internal/wrphandlers/qos/priority_queue.go
johnabass: I would have a max message size as well.  It's generally a best practice to bound both (1) the number of units of work, and (2) the size of a single unit of work.

That does leave you with a design decision though:  What do you do with messages that are too large?  How do you indicate that to the caller?  There's lots of right ways to do that.

internal/wrphandlers/qos/qos.go
johnabass: It's fine to leave a channel open.  It's not a channel in the IO sense of the word.  It's just a blocking queue.  Closing a `golang` channel is just the equivalent of sending a sentinel value to receivers that no further data on the channel is coming.

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,138 @@
...
+		if opt != nil {
+			if err := opt.apply(h); err != nil {
+				return nil, nil, err
+			}
+		}
+	}
+
+	// shutdown() is used to stop serviceQOS by closing its `done` chan.
+	ctx, shutdown := context.WithCancel(context.Background())
+	go serviceQOS(ctx.Done(), h.queue, h.maxHeapSize, handleWRPWrapper(next))
```
johnabass: Starting a goroutine here may cause issues with `fx`.  This may cause the constructor `New` to take longer than `fx` wants to wait, especially in a constrained environment.  It might be better to bind the lifecycle of a `Handler` to the enclosing `fx.App`.

internal/wrphandlers/qos/qos.go
denopink: I didn't know that, good to know!

internal/wrphandlers/qos/qos.go
denopink: that makes sense, I'll go ahead and re-introduce the `Start` and `Stop` methods and add some atomic state management around starting/stopping the qos goroutine

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -0,0 +1,99 @@
...
+		return wrp.Message{}, false
+	}
+
+	msg, ok := heap.Pop(pq).(wrp.Message)
+
+	return msg, ok
+}
+
+// Enqueue queues the given message.
+func (pq *priorityQueue) Enqueue(msg wrp.Message) {
```
denopink, line 34: if we wanted to enforce a max message payload, then what size limit would we start with and do we want it to be configurable?

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,191 @@
...
+func (h *Handler) Start() error {
+	h.lock.Lock()
+	defer h.lock.Unlock()
+
+	if h.ctx != nil {
+		return nil
+	}
+
+	// h.queue will never block as long as the serviceQOS goroutine is running.
+	h.queue = make(chan wrp.Message, 1)
```
denopink: technically `h.queue` will never block as long as the serviceQOS goroutine is running... is that enough insurance or should add more of a buffer to be extra safe?

internal/wrphandlers/qos/priority_queue.go
schmidtw: I don't think this is the handler is the place to enforce anything about the wrp message.  This should focus only on applying qos rules.

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -0,0 +1,99 @@
...
+	// Check whether msg would already violate maxQueueSize.
+	if len(msg.Payload) > pq.maxQueueSize {
+		return
+	}
+
+	// Check whether enqueuing msg would violate maxQueueSize.
```
schmidtw, line 40: I'd add the new message, then create a new function:
```golang
func (pq *priorityQueue) drop() {
    _ = heap.Remove(pq, pq.Len()-1)
}
```

This makes the code in the for loop much simpler to read and understand because you'll just loop until the size is less than or equal to the max size.

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,190 @@
...
+	lock   sync.Mutex
+	ctx    context.Context
+	cancel context.CancelFunc
+}
+
+// New creates a new instance of the Handler struct.  The parameter next is the
+// handler that will be called and monitored for errors.
+// Note, once cancel is called, any calls to Handler.HandleWrp will result in
+// an ErrQOSHasShutdown error
+func New(next websocket.Egress, opts ...Option) (h *Handler, err error) {
```
schmidtw: Why is next `websocket.Egress` and not a `wrpkit.Handler`?

internal/wrphandlers/qos/qos.go
schmidtw: That's not exactly true.  It depends on when the queue gets a chance to be read from.  Blocking producers isn't really a problem generally.  One is fine, zero may be fine, too.

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	return h, errs
+}
+
+func (h *Handler) Start() error {
+	h.lock.Lock()
+	defer h.lock.Unlock()
+
+	if h.ctx != nil {
```
schmidtw, line 77: Switch this to check if the `h.queue` is `nil` instead.

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,190 @@
...
+
+		return failedMsg, ready
+	}
+}
+
+// serviceQOS is a long running goroutine that sends as many queued messages as possible,
+// where the highest QOS messages are prioritized.
+// serviceQOS starts when Handler.Start() is called.
+// serviceQOS stops when Handler.Stop() is called, closing its `done` chan.
+func serviceQOS(done <-chan struct{}, queue <-chan wrp.Message, maxQueueSize int, handleWRP serviceQOSHandler) {
```
schmidtw, line 144: This should be a method on the handler, not an unrelated function.

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,190 @@
...
+
+	if h.ctx != nil {
+		return nil
+	}
+
+	// h.queue will never block as long as the serviceQOS goroutine is running.
+	h.queue = make(chan wrp.Message, 1)
+	// h.cancel() stops serviceQOS by closing its `done` chan.
+	h.ctx, h.cancel = context.WithCancel(context.Background())
+	go serviceQOS(h.ctx.Done(), h.queue, h.maxQueueSize, h.next)
```
schmidtw, line 84: I'd change `serviceQOS` to a method like `func (h *Handler) run()`.  It's simpler & what folks expect to see.

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,190 @@
...
+// Handler queues incoming messages and sends them to the next wrphandler
+type Handler struct {
+	next serviceQOSHandler
+	// queue for wrp messages, ingested by serviceQOS
+	queue chan wrp.Message
+	// maxQueueSize is the allowable max size of the qos' priority queue, based on the sum of all queued wrp message's payload
+	maxQueueSize int
+
+	lock   sync.Mutex
+	ctx    context.Context
```
schmidtw, line 43: By using a `queue` above, you don't need a context or cancel function.  Why?  The goroutine runs when the queue is around.  Close the queue and the routine is done.  Don't duplicate signals if at all possible.  It doesn't make things more reliable, it actually makes them less.

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,190 @@
...
+	go serviceQOS(h.ctx.Done(), h.queue, h.maxQueueSize, h.next)
+
+	return nil
+}
+
+func (h *Handler) Stop() {
+	h.lock.Lock()
+	defer h.lock.Unlock()
+
+	if h.ctx == nil {
```
schmidtw, line 93: The rest of this function can be:
```golang
if h.queue != nil {
    close(h.queue)
    h.queue = nil
}
```

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,190 @@
...
+// to send as many queued messages as possible, where the highest QOS messages are prioritized
+func (h *Handler) HandleWrp(msg wrp.Message) error {
+	h.lock.Lock()
+	defer h.lock.Unlock()
+
+	if h.queue == nil {
+		return ErrQOSHasShutdown
+	}
+
+	// queue will never block as long as the serviceQOS goroutine is running.
```
schmidtw: I'd remove the comment.  It's not entirely true.

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,190 @@
...
+	h.queue <- msg
+
+	return nil
+}
+
+func handleWRPWrapper(next wrpkit.Handler) serviceQOSHandler {
```
schmidtw, line 120: This seems like an elaborate way of simply calling `err = next.HandleWrp(msg)` inline ... why do we need to do this?

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,190 @@
...
+				return
+			}
+
+			pq.Enqueue(msg)
+			if ready != nil {
+				// Previous handleWRP call has not finished, do nothing.
+			} else if top, ok := pq.Dequeue(); ok {
+				failedMsg, ready = handleWRP(top)
+			}
+		case <-ready:
```
schmidtw, line 171: This is unnecessarily complex.  Just select on the queue, call `next.HandleWrp()` block the goroutine and handle the results.  Then you can delete this entire block of code.

schmidtw (changes requested):

internal/wrphandlers/qos/priority_queue.go
denopink: that's a great point!

internal/wrphandlers/qos/qos.go
denopink: good catch, I forgot to switch to `wrpkit.Handler`

internal/wrphandlers/qos/priority_queue.go
denopink: agree 🙂 

From our side conversation it sounded like we're considering making this a configurable feature, right?
I.e. `should we favor old or new messages [when the queue is full]?`

- If `we favor old messages`, then we want to check whether the oldest message is of higher priority than the new message - keeping the old and dropping the new message.
- If `we favor new messages`, then we always enqueue the new message while dropping the oldest message.

internal/wrphandlers/qos/qos.go
denopink: I think I understand the reasoning behind your suggestion, since qos's `HandleWrp()` is already being called in this goroutine `err := h.HandleWrp(*normalized)`
https://github.com/xmidt-org/xmidt-agent/blob/78dffe0cad394ab82f581940e3a8117f04941077/internal/pubsub/pubsub.go#L196-L211

then whether qos's `HandleWrp()` blocks  or not doesn't matter since it's not blocking anything - therefore, we can simplify things by removing this, right? Or was there another reason behind this suggestion?

I'm cool with this idea

internal/wrphandlers/qos/qos.go
denopink: that totally makes sense, why keep track of more things when you don't have to 🙂

internal/wrphandlers/qos/qos.go
denopink: it was handy for john's approach https://github.com/xmidt-org/xmidt-agent/pull/104#discussion_r1572879127

but your suggestion https://github.com/xmidt-org/xmidt-agent/pull/104/files#r1585910687 seems simpler and gets the job done

denopink:
continuing work on #114

PR #106, opened 2024-04-19
add ability to dynamically adjust loglevel

piccione99:
This PR adds an XmidtAgent specific crud service to listen to commands.  The only currently implemented command is logLevel and it is an update message. 

The following assumptions are made and maybe be wrong so let's verify before merging:

1. dest of the wrp message will route it to xmidtAgentCrud 
2. path of the wrp.message specifies what we are updating (in this case "loglevel")


cmd/xmidt-agent/wrphandlers.go
```diff
@@ -92,6 +95,33 @@ func provideAuthHandler(in authIn) (*auth.Handler, error) {
...
+}
+
+func provideCrudHandler(in crudIn) (*xmidt_agent_crud.Handler, error) {
+	h, err := xmidt_agent_crud.New(in.Egress, string(in.Identity.DeviceID), in.LogLevelService)
+	if err != nil {
+		err = errors.Join(ErrWRPHandlerConfig, err)
+		return nil, err
+	}
+
+	// what do we do with the return value?  Does this have to be passed to pubSub somehow? Seems
```
schmidtw, line 115: It's safe to ignore it.  Only if you need the ability to cancel the subscription does it matter.

internal/loglevel/loglevel.go
```diff
@@ -0,0 +1,51 @@
+// SPDX-FileCopyrightText: 2023 Comcast Cable Communications Management, LLC
```
schmidtw: It's 2024.

internal/loglevel/loglevel_test.go
```diff
@@ -0,0 +1,43 @@
+// SPDX-FileCopyrightText: 2023 Comcast Cable Communications Management, LLC
```
schmidtw: It's 2024.

internal/wrphandlers/xmidt_agent_crud/handler.go
```diff
@@ -0,0 +1,103 @@
...
+	err := json.Unmarshal(msg.Payload, &payload)
+	if err != nil { // do we still want to return a response here?
+		return err // this would be sent to the eventor error handler
+	}
+
+	var payloadResponse []byte
+	statusCode := int64(http.StatusOK)
+
+	switch msg.Type {
+	case Update:
```
schmidtw: Use `wrp.UpdateMessageType` instead of defining duplicates in this file.

internal/wrphandlers/xmidt_agent_crud/handler.go
```diff
@@ -0,0 +1,103 @@
...
+		statusCode, _ = h.update(msg.Path, payload)
+		// the above error needs to be logged
```
schmidtw: Set the payload response to be the err if present, so it is sent to the original caller.

internal/wrphandlers/xmidt_agent_crud/handler.go
```diff
@@ -0,0 +1,103 @@
...
+func (h *Handler) HandleWrp(msg wrp.Message) error {
+	payload := make(map[string]interface{})
+
+	err := json.Unmarshal(msg.Payload, &payload)
+	if err != nil { // do we still want to return a response here?
+		return err // this would be sent to the eventor error handler
+	}
+
+	var payloadResponse []byte
+	statusCode := int64(http.StatusOK)
```
schmidtw: Unless recognized and correct, `http.StatusOK` should probably be `http.StatusBadRequest`

internal/wrphandlers/xmidt_agent_crud/handler.go
```diff
@@ -0,0 +1,103 @@
...
+	response.Payload = payloadResponse
+
+	response.Status = &statusCode
+
+	err = h.egress.HandleWrp(response)
+
+	return err
+}
+
+// why is status code int64?
```
schmidtw: I think it's a relic of ugorji.

internal/wrphandlers/xmidt_agent_crud/handler.go
```diff
@@ -0,0 +1,103 @@
...
+func (h *Handler) update(path string, payload map[string]interface{}) (int64, error) {
+	switch path {
+	case "loglevel":
+		err := h.changeLogLevel(payload)
+		if err != nil {
+			return int64(http.StatusBadRequest), err
+		}
+
+	default:
+		return int64(http.StatusOK), nil
```
schmidtw: I'd only return 200 when it's known and handled.

internal/wrphandlers/xmidt_agent_crud/handler.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	return int64(http.StatusOK), nil
+}
+
+func (h *Handler) changeLogLevel(payload map[string]interface{}) error {
+	minutes := float64(DefaultLogLevelChangeMinutes)
+	inputMinutes := payload["duration"]
```
schmidtw: Use the `time.Duration` type instead of minutes.

schmidtw (changes requested):

internal/loglevel/loglevel.go
piccione99: I'll update all the other files in a separate commit.  most of the repo is 2023

PR #108, opened 2024-04-23
update dockerfile and default config 

piccione99:
1. updated default config so we don't need a config file for running docker
2. updated readme instructions
3. added an example config file for reference 
4. couldn't get entrypoint.sh to work so currently not using it


Dockerfile
```diff
@@ -27,12 +28,16 @@ COPY LICENSE    /
...
-ENTRYPOINT ["/entrypoint.sh"]
+#ENTRYPOINT ["tail", "-f", "/dev/null"]
+#ENTRYPOINT ["/entrypoint.sh"]
+ENTRYPOINT ["/xmidt-agent"]
 
 EXPOSE 6666
 
-CMD ["/xmidt-agent"]
\ No newline at end of file
+CMD ["-f", "/etc/xmidt-agent/xmidt_agent.yaml", "--dev"]
```
piccione99: TODO - this doesn't work without specifying the yaml file, shouldn't need to

ilawjr (approved):

denopink (approved):
the idea makes sense since it's a pattern we already put into practice for our services, i.e.: the service should be able to start with the default config.

However, do we want that to be true for xmidt-agent since it's running on a cpe? Not sure if that's a desired pattern for software running on cpe.

Approving this, but I think we shouldn't merged until @schmidtw chimes in.

schmidtw:
Yes, in general we want the app to basically work right out of the box.  We also want it to be secure out of the box, too.  I will review this later; there are a few parts being built into the configuration that I think we should do via environment variables (the `device_id` and other identity items being at the top of this list).

PR #111, opened 2024-04-29
patch:  simplify fx injection with `[]func() `group:"cancels"``

### LINKED ISSUE #109: simplify fx injection by replacing the types `wrp.HandlerFunc` and websocket's `event.CancelFunc` with `func()`
replacing the types`wrp.HandlerFunc` and websocket's `event.CancelFunc` with `func()`   would allow us to use fx value groups and simplify our lifecycle's onStop. i.e. replace `eventCancelList` and `pubsubCancelList` with `cancels` (a list of` func()`s) 

https://github.com/xmidt-org/xmidt-agent/blob/4eefba0f09d888a6ec1f8db2519e2645b641dbd2/cmd/xmidt-agent/main.go#L256-L263

denopink:
implements #109 

- Implicitly cast `CancelFunc` to `func()`.
- patch `func (ws *Websocket) AddMessageListener`

### DIFF
--- a/internal/websocket/options.go
+++ b/internal/websocket/options.go
@@ -87,6 +87,18 @@ func CredentialsDecorator(f func(http.Header) error) Option {
 		})
 }
 
+func ConveyDecorator(f func(http.Header) error) Option {
+	return optionFunc(
+		func(ws *Websocket) error {
+			if f == nil {
+				return fmt.Errorf("%w: nil ConveyDecorator", ErrMisconfiguredWS)
+			}
+
+			ws.conveyDecorator = f
+			return nil
+		})
+}
+
 // PingInterval sets the time expected between PINGs for the WS connection.
 // If this is not set, the default is 30 seconds.
 func PingInterval(d time.Duration) Option {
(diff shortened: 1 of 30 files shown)


internal/websocket/options.go
```diff
@@ -273,45 +273,45 @@ func MaxMessageBytes(bytes int64) Option {
 
 // AddMessageListener adds a message listener to the WS connection.
 // The listener will be called for every message received from the WS.
-func AddMessageListener(listener event.MsgListener, cancel ...*event.CancelFunc) Option {
+func AddMessageListener(listener event.MsgListener, cancel ...*func()) Option {
```
schmidtw: I don't think this is the right way to achieve this.  Can we not explicitly cast the result to a `func()` afterwards?

denopink: for sure! actually we don't need too since it'll be implicitly casted.

replacing our `CancelFunc` with `func()` was inspired by your eventor update https://github.com/xmidt-org/eventor/pull/7/files

schmidtw: Ok, I see the trend is to be more consistent.  Great!

internal/websocket/ws.go
```diff
@@ -187,7 +187,7 @@ func (ws *Websocket) HandleWrp(m wrp.Message) error {
 
 // AddMessageListener adds a message listener to the WS connection.
 // The listener will be called for every message received from the WS.
-func (ws *Websocket) AddMessageListener(listener event.MsgListener, cancel ...*event.CancelFunc) event.CancelFunc {
```
denopink, line 190: I meant to remove `cancel ...*event.CancelFunc` when we integrate the wrphandlers

schmidtw (approved):

PR #114, opened 2024-05-03
feat: qos integration

denopink:
100% test coverage


cmd/xmidt-agent/config.go
```diff
@@ -37,10 +37,18 @@ type Config struct {
...
 	Storage          Storage
 	MockTr181        MockTr181
+	QOS              QOS
 	Externals        []configuration.External
 	XmidtAgentCrud   XmidtAgentCrud
 }
 
+type QOS struct {
+	// MaxQueueSize is the allowable max size of the qos' priority queue, based on the sum of all queued wrp message's payload.
+	MaxQueueSize int
```
schmidtw: I'd rename this `MaxQueueBytes` to be consistent with `MaxMessageBytes`

cmd/xmidt-agent/wrphandlers.go
```diff
@@ -132,7 +148,7 @@ type pubsubIn struct {
 	MockTr181 MockTr181
 
 	// wrphandlers
-	Egress websocket.Egress
+	Egress *qos.Handler
```
schmidtw, line 152: Doesn't the `missing.Handler` need to be called before QOS?

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -0,0 +1,101 @@
...
+	msg, ok := heap.Pop(pq).(wrp.Message)
+
+	return msg, ok
+}
+
+// Enqueue queues the given message.
+func (pq *priorityQueue) Enqueue(msg wrp.Message) {
+	// Check whether msg violates maxMessageBytes.
+	if len(msg.Payload) > pq.maxMessageBytes {
+		return
```
schmidtw: Dropping the incoming message because it's too large should return an error.  This is different than the message (or any message) being dropped due to being lowest priority.

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	// Check whether enqueuing msg would violate maxQueueSize.
```
schmidtw: I'd simply add & then shrink afterwards until the size is acceptable to remove duplicate code.

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -0,0 +1,101 @@
...
+		if bottom.QualityOfService > msg.QualityOfService {
+			msg = bottom
+		}
+	}
+
+	heap.Push(pq, msg)
+}
+
+func (pq *priorityQueue) Size() int {
+	var s int
```
schmidtw: Since Size() is looped over each insertion, I'd favor a field of `size` that is updated in `Push()` and `Pop()`.

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,194 @@
...
+// Note, once cancel is called, any calls to Handler.HandleWrp will result in
+// an ErrQOSHasShutdown error
+func New(next wrpkit.Handler, opts ...Option) (h *Handler, err error) {
+	if next == nil {
+		return nil, ErrInvalidInput
+	}
+
+	opts = append(opts, validateQueueConstraints())
+
+	h = &Handler{
```
schmidtw: Favor invoking a non-pointer type & pass in the pointer when it is needed.

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,194 @@
...
+	defer h.lock.Unlock()
+
+	if h.ctx != nil {
+		return nil
+	}
+
+	h.queue = make(chan wrp.Message)
+	// h.cancel() stops serviceQOS by closing its `done` chan.
+	h.ctx, h.cancel = context.WithCancel(context.Background())
+	go serviceQOS(h.ctx.Done(), h.queue, h.maxQueueSize, h.maxMessageBytes, h.next)
```
schmidtw: Make `serviceQOS()` a method & don't pass the fields from the object.  Pass the context, not the Done.  That way if the function needs a sub context it can create and link one.

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	return h, errs
+}
+
+func (h *Handler) Start() error {
+	h.lock.Lock()
+	defer h.lock.Unlock()
+
+	if h.ctx != nil {
```
schmidtw: Only keep the cancel function & check against it.

schmidtw (changes requested):

internal/wrphandlers/qos/qos.go
denopink: based on our side conversation, I'm assuming we still want move forward without the context/cancel

internal/wrphandlers/qos/qos.go
```diff
@@ -0,0 +1,192 @@
...
+	// handleWRP is promise pattern function
+	handleWRP := curryWRPHandler(h.next)
+
+	h.lock.Lock()
+	queue := h.queue
+	h.lock.Unlock()
+
+	if queue == nil {
+		return
+	}
```
denopink, line 135: did not mean to commit this

```suggestion
	queue := h.queue
	if queue == nil {
		return
	}
```

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -45,20 +46,18 @@ func (pq *priorityQueue) Enqueue(msg wrp.Message) error {
...
+func (pq *priorityQueue) trim() {
+	// trim until the queue no longer violates maxQueueBytes.
+	for pq.sizeBytes > pq.maxQueueBytes {
+		// Note, priorityQueue.drop does not drop the least prioritized queued message.
+		// i.e.: a high priority queued message may be dropped instead of a lesser queued message.
 		pq.drop()
 	}
-
-	return nil
 }
```
denopink, line 61: I couldn't move the trimming code to `priorityQueue.Push` without extra code with complicated swapping

Trimming after `heap.Push`'s `up(h, h.Len()-1)` makes things simpler imo 
https://github.com/golang/go/blob/4ed358b57efdad9ed710be7f4fc51495a7620ce2/src/container/heap/heap.go#L51-L54
```go
// Push pushes the element x onto the heap.
// The complexity is O(log n) where n = h.Len().
func Push(h Interface, x any) {
	h.Push(x)
	up(h, h.Len()-1)
}
```

internal/wrphandlers/qos/qos.go
```diff
@@ -111,79 +109,62 @@ func (h *Handler) HandleWrp(msg wrp.Message) error {
...
 			}
 
-			pq.Enqueue(msg)
-			if ready != nil {
-				// Previous handleWRP call has not finished, do nothing.
-			} else if top, ok := pq.Dequeue(); ok {
-				failedMsg, ready = handleWRP(top)
-			}
+			// ErrMaxMessageBytes errrors are ignored.
+			_ = pq.Enqueue(msg)
```
denopink, line 133: if a ErrMaxMessageBytes error occurs, should qos still send a message (with a nil payload) if the message requires a transaction?

If so, what status code should we use?

internal/wrphandlers/qos/priority_queue.go
schmidtw: That is reasonable.

schmidtw (approved):

PR #115, opened 2024-05-03
add configuarable convey header 

piccione99:
Note that this does NOT handle webpa-interface-used yet.  This will be a separate story and might require some small amount of refactoring because interface used most likely needs to be determined by the websocket.


internal/convey/convey.go
```diff
@@ -0,0 +1,115 @@
...
+		}
+	}
+
+	return header
+}
+
+func (c *ConveyHeaderProvider) Decorate(headers http.Header) error {
+	header := c.GetConveyHeader()
+	headerBytes, err := json.Marshal(header)
+	if err != nil {
```
schmidtw: I think the error handling needs to either fail on error, or allow configuration to silently fail on error (if that is wanted... I'm not sure this is desired).

I'm not sure how marshal can fail that we'd want to accept since it's exclusively an error focused on exporting the existing configuration.

internal/convey/convey.go
```diff
@@ -0,0 +1,115 @@
...
+			if err := opt.apply(conveyHeaderProvider); err != nil {
+				return nil, err
+			}
+		}
+	}
+
+	return conveyHeaderProvider, nil
+}
+
+func (c *ConveyHeaderProvider) GetConveyHeader() map[string]interface{} {
```
schmidtw: Instead of getting the header, I'd just make the ConveyHeaderProvider have a few decorators (one for HTTP and one for WRP)

schmidtw (changes requested):

internal/convey/convey.go
```diff
@@ -0,0 +1,115 @@
...
+	SerialNumber               = "hw-serial-number"
+	LastRebootReason           = "hw-last-reboot-reason"
+	Protocol                   = "webpa-protocol"
+	BootTime                   = "boot-time"
+	BootTimeRetryDelay         = "boot-time-retry-wait"
+	InterfaceUsed       string = "webpa-interface-used"
+	InterfacesAvailable        = "interfaces-available"
+)
+
+type ConveyHeaderProvider struct {
```
schmidtw: Overall this is good code.  I wonder if this should be something like `MetadataProvider` that accepts configuration for Convey as well as for WRP messages.  The code below can provide for Convey & that decorator & maybe add a 2nd configuration for WRPMetaDataAdded that is basically the same list of options & a decorator that decorates the WRP metadata.

My thinking on this is if we have one place that collects the metadata (you've done that here) and it is used in both places then the data should always be consistent.  ... or more consistent.

schmidtw:
Overall this is good code. I wonder if this should be something like `MetadataProvider` that accepts configuration for `Convey` as well as for WRP messages. The code below can provide for Convey & that decorator & maybe add a 2nd configuration for WRPMetaDataAdded that is basically the same list of options & a decorator that decorates the WRP metadata.

My thinking on this is if we have one place that collects the metadata (you've done that here) and it is used in both places then the data should always be consistent. ... or more consistent.

internal/convey/convey.go
piccione99: I want to keep that function so I can use it for both decorators

internal/convey/convey.go
piccione99: oversight, returning an error now if marshal fails

schmidtw (approved):

PR #121, opened 2024-05-10
feat: add http client configuration to websocket

### LINKED ISSUE #66: Add http client configuration to websocket.

denopink:
implements #66

- Leverages arrangehttp.ClientConfig as the configuration struct and constructor for the websocket's http client.
- Instantiate a new http client for every connection attempt.


internal/websocket/options.go
```diff
@@ -218,16 +178,51 @@ func SendTimeout(d time.Duration) Option {
 		})
 }
 
-// ConnectTimeout sets the timeout for the WS connection.  If this is not set,
-// the default is 30 seconds.
-func ConnectTimeout(d time.Duration) Option {
+// HTTPClient is the configuration for the HTTP client used to retrieve the
+// credentials.
+func HTTPClient(c arrangehttp.ClientConfig) Option {
```
schmidtw: This needs to be an optional Option.  Right now it is required since it creates the http.Client used.  I think this is a mistake & that this should largely just take a `*http.Client` in as input.  Later (after the options are handled) the final `*http.Client` should be created given the need for the `ipv4` vs `ipv6` choice as well as which network interface to attempt to use.

internal/websocket/ws.go
```diff
@@ -330,14 +321,11 @@ func (ws *Websocket) dial(ctx context.Context, mode ipMode) (*nhws.Conn, *http.R
 	if err != nil {
 		return nil, nil, err
 	}
-
+	ws.updateTransportDialContext(mode)
 	conn, resp, err := nhws.Dial(ctx, url,
```
schmidtw, line 327: This is the place the the final `*http.Client` should be created.  It should not be stored, just passed to the `Dial()` function.

schmidtw (changes requested):

internal/websocket/options.go
denopink: great point(!), I updated it such that `func HTTPClient(c arrangehttp.ClientConfig)` is actually optional and the http client is created as needed 🙂

internal/websocket/options.go
```diff
@@ -218,16 +178,38 @@ func SendTimeout(d time.Duration) Option {
...
+			}
+
+			if errs != nil {
+				return errs
+			}
+
+			// test new client
+			_, err = c.NewClient()
+			if err != nil {
+				return errors.Join(errs, err)
```
denopink: typo
```suggestion
				return errors.Join(ErrMisconfiguredWS, err)
```

internal/websocket/options.go
```diff
@@ -218,16 +178,38 @@ func SendTimeout(d time.Duration) Option {
 		})
 }
 
-// ConnectTimeout sets the timeout for the WS connection.  If this is not set,
-// the default is 30 seconds.
-func ConnectTimeout(d time.Duration) Option {
+// HTTPClient is the configuration for the HTTP client used for connection attempts.
+func HTTPClient(c arrangehttp.ClientConfig) Option {
```
schmidtw: Instead of passing the `arrangehttp.ClientConfig` pass the `*http.Client` value here.  This is consistent with other similar interfaces & decouples configuration from use a bit more.  Also I don't think we need to boundary check the client since we're being told what someone wants & that will remove some code as well.

schmidtw (changes requested):

schmidtw (approved):

PR #122, opened 2024-05-13
add stub for setting and retrieving interface-used from the websocket

piccione99:
The actual setting of interface used is not implemented here because it requires a custom net.Dialer which is still outstanding in a different commit. 

This just stubs out how to get the interface-used value from websocket to the metadata provider. 

**** Note that "allowedInterfaces" is configurable.  This seems shaky if network interface names change. ***

### DIFF
--- a/cmd/xmidt-agent/config.go
+++ b/cmd/xmidt-agent/config.go
@@ -22,6 +22,7 @@ import (
 	"github.com/xmidt-org/sallust"
 	"github.com/xmidt-org/wrp-go/v3"
 	"github.com/xmidt-org/xmidt-agent/internal/configuration"
+	"github.com/xmidt-org/xmidt-agent/internal/net"
 	"go.uber.org/zap/zapcore"
 	"gopkg.in/dealancer/validate.v2"
 )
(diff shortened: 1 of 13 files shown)


cmd/xmidt-agent/config.go
```diff
@@ -397,4 +403,7 @@ var defaultConfig = Config{
 	Metadata: Metadata{
 		Fields: []string{"fw-name", "hw-model", "hw-manufacturer", "hw-serial-number", "hw-last-reboot-reason", "webpa-protocol", "boot-time", "boot-time-retry-wait", "webpa-interface-used", "interfaces-available"},
 	},
+	NetworkService: NetworkService{
+		AllowedInterfaces: []string{"erouter0", "eroutev0", "br-home", "brrwan", "vdsl0", "wwan0", "wlan0", "eth0", "qmapmux0.127", "cm0"},
```
piccione99: this configuration seems shaky if network interface names change

denopink (approved):
do we plan to integrate this to the websocket in another pr?

Otherwise, lgtm! 🍻

PR #124, opened 2024-05-15
patch: fx lifecycle panics should return an error while rolling back and shuting down

denopink:
#133 must be merged first
 
- Currently if a panic occurs during our fx lifecycle, we log the panic and then ignore it.
- Instead, we want to return an error in order for fx to rollback and shutdown

### DIFF
--- a/cmd/xmidt-agent/main.go
+++ b/cmd/xmidt-agent/main.go
@@ -5,6 +5,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"runtime/debug"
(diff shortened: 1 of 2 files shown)


denopink (commented):
I'll write a test for this.

cmd/xmidt-agent/main.go
```diff
@@ -59,6 +66,16 @@ type LifeCycleIn struct {
...
+	app := fx.New(provideAppOptions(args))
+	if err := app.Err(); err != nil {
+		return nil, err
+	}
+
+	return app, nil
+}
+
+// provideAppOptions returns all fx options required to start the xmidt agent fx app.
+func provideAppOptions(args []string) fx.Option {
```
denopink, line 78: `provideAppOptions` helps simplify lifecycle testing

cmd/xmidt-agent/main_test.go
```diff
@@ -14,6 +15,9 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/xmidt-org/sallust"
+	"github.com/xmidt-org/xmidt-agent/internal/websocket"
+	"github.com/xmidt-org/xmidt-agent/internal/wrphandlers/qos"
+	"go.uber.org/fx"
 )
```
denopink, line 21: Note, `Test_xmidtAgent()` will fail until #133 is merged

schmidtw (approved):

PR #125, opened 2024-05-15
feat: support openfail by default

denopink:
implements openfail

- if the agent's config is missing the `xmidt_credentials`, then the agent will connect to xmidt via openfail

### DIFF
--- a/cmd/xmidt-agent/ws.go
+++ b/cmd/xmidt-agent/ws.go
@@ -6,7 +6,6 @@ package main
 import (
 	"context"
 	"errors"
-	"net/http"
 	"net/url"
 	"time"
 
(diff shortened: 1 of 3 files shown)


cmd/xmidt-agent/ws.go
```diff
@@ -58,6 +59,13 @@ func provideWS(in wsIn) (wsOut, error) {
 		fetchURLFunc = in.JWTXT.Endpoint
 	}
 
+	credDecorator := func(http.Header) error { return nil }
```
denopink: I'll simplify this when https://github.com/xmidt-org/xmidt-agent/pull/123 is merged

piccione99 (commented):
looks good to me.  the real production servers will reject attempts for xmidt-agent to connect?

piccione99 (approved):

schmidtw (approved):

PR #129, opened 2024-05-16
refactor: Refactor pubsub package in order to keep defaults consistent

denopink:
- Refactor the pubsub package in order to keep defaults consistent, since defaults are set in `cmd/xmidt-agent/config.go`

### DIFF
--- a/cmd/xmidt-agent/config.go
+++ b/cmd/xmidt-agent/config.go
@@ -362,7 +362,7 @@ var defaultConfig = Config{
 		},
 	},
 	Pubsub: Pubsub{
-		PublishTimeout: 200 * time.Millisecond,
+		PublishTimeout: 5 * time.Second,
 	},
 	Logger: sallust.Config{
 		EncoderConfig: sallust.EncoderConfig{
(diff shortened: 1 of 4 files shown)


internal/pubsub/pubsub.go
```diff
@@ -69,13 +69,6 @@ func New(self wrp.DeviceID, opts ...Option) (*PubSub, error) {
 		),
 	}
 
-	defaults := []Option{
-		WithPublishTimeout(5 * time.Second),
-	}
-
-	// Prepend the defaults to the provided options.
-	opts = append(defaults, opts...)
-
```
denopink, line 78: removing in order to keep things consistent (where our defaults are set), since we already set defaults in `cmd/xmidt-agent/config.go`

cmd/xmidt-agent/config.go
```diff
@@ -362,7 +362,7 @@ var defaultConfig = Config{
 		},
 	},
 	Pubsub: Pubsub{
-		PublishTimeout: 200 * time.Millisecond,
+		PublishTimeout: 5 * time.Second,
```
denopink, line 372: use the actual pubsub package default `5 * time.Second`

schmidtw (approved):

PR #131, opened 2024-05-16
patch: reused http client will trigger a panic during `updateClientTransport()`

denopink:
- when `updateClientTransport()` is called more than once, `ws.client.Transport.(*http.Transport)` will trigger the panic `http.RoundTripper is *websocket.custRT, not *http.Transport`

### DIFF
--- a/internal/websocket/ws.go
+++ b/internal/websocket/ws.go
@@ -369,17 +369,32 @@ func (rt *custRT) RoundTrip(r *http.Request) (*http.Response, error) {
 // updateClientTransport updates the http client's Transport and set the DialContext's
 // named network as the provided `mode`.
 func (ws *Websocket) updateClientTransport(mode ipMode) {
+	var (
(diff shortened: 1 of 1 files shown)


internal/websocket/ws.go
```diff
@@ -371,17 +371,42 @@ func (rt *custRT) RoundTrip(r *http.Request) (*http.Response, error) {
 func (ws *Websocket) updateClientTransport(mode ipMode) {
 	// Override client's Transport with custRT (reusing certain configurations)
 	// and update it's DialContext with the provided mode.
```
denopink: when `updateClientTransport()` is called more than once, `ws.client.Transport.(*http.Transport)` will trigger the panic `http.RoundTripper is *websocket.custRT, not *http.Transport`

denopink:
closing since we're reverting back to https://github.com/xmidt-org/xmidt-agent/commit/254f9b98811a06940393667398af0e26c9bcb6b2

PR #128, opened 2024-05-16
feat: integrate libparodus

denopink:
integrates libparodus

PR #134, opened 2024-05-17
patch: update tests main_test and local default config

denopink:
- tests broke when #128 merged
- update docs
- https://github.com/xmidt-org/xmidt-agent/commit/1acb340ae73a86862edaab9ce0736b556bac89a6 fixed another issue created by #128, but the new config section `externals` caused main_test tests to fail since `/home/schmidtw/sync/work/oss/xmidt-agent/main/xmidt-agent.main/cmd/xmidt-agent/wes.txt` didn't exist.
- we're getting rid of our centralize lifecycle functions in main, so we're removing the lifecycle tests (the lifecycle tests were failing because multiple libparodus were trying to use  the same port)

### DIFF
--- a/cmd/xmidt-agent/main_test.go
+++ b/cmd/xmidt-agent/main_test.go
@@ -5,7 +5,6 @@ package main
 
 import (
 	"context"
-	"errors"
 	"testing"
 	"time"
 
@@ -15,9 +14,6 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/xmidt-org/sallust"
-	"github.com/xmidt-org/xmidt-agent/internal/websocket"
-	"github.com/xmidt-org/xmidt-agent/internal/wrphandlers/qos"
-	"go.uber.org/fx"
 )
 
 func Test_provideCLI(t *testing.T) {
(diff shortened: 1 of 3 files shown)


cmd/xmidt-agent/main_test.go
```diff
@@ -158,12 +158,12 @@ func Test_xmidtAgent_lifecycle(t *testing.T) {
 			description: "panic in fx's Start triggered a successful rollback",
 			lifeCycleOptions: fx.Invoke(
 				func(LC fx.Lifecycle, ws *websocket.Websocket) {
-					var qos *qos.Handler
```
denopink, line 161: tests broke when #128 merged

cmd/xmidt-agent/xmidt_agent.yaml
```diff
@@ -26,22 +26,22 @@ operational_state:
...
+# Optional (Examples)
 # storage:
 #   temporary: ~/local-rdk-testing/temporary
 #   durable: ~/local-rdk-testing/durable
 # mock_tr_181:
 #   enabled: true
 #   file_path: xmidt-agent/internal/wrphandlers/mocktr181/mock_tr181.json
-externals:
-  -
-    file: /home/schmidtw/sync/work/oss/xmidt-agent/main/xmidt-agent.main/cmd/xmidt-agent/wes.txt
```
denopink, line 38: i figured we meant to comment this out or remove it 🙂

schmidtw: This was added as a mistake & should be here at all :crying_cat_face:

cmd/xmidt-agent/main_test.go
```diff
@@ -141,93 +137,6 @@ func Test_xmidtAgent(t *testing.T) {
 	}
 }
 
-type badShutdown struct{}
-
-func (*badShutdown) Shutdown(...fx.ShutdownOption) error {
-	return errors.New("random shutdown error")
-}
-
-func Test_xmidtAgent_lifecycle(t *testing.T) {
```
denopink, line 150: FYI we're getting rid of our centralize lifecycle functions in main, so we wont need these tests (the lifecycle tests were failing because multiple libparodus were trying to use  the same port)

schmidtw (approved):

PR #140, opened 2024-05-18
feat: implement websocket inactivity timeout

### LINKED ISSUE #119: feat: disconnect & reconnect when pings are not received within the configured ping time interval (`pingInterval`)
This is configured with `pingInterval` and it's currently not in use:
https://github.com/xmidt-org/xmidt-agent/blob/772f52e3363e5c8fd40de02da85fbb85d5547877/internal/websocket/ws.go#L50-L51

denopink:
The main focus of this pr is in `internal/nhooyr.io/websocket/conn.go`:
- ~~Implemented it in the [nhooyr.io/websocket] package rather than in our `websocket` package, in order to minimize edits to existing code and I think it it fits better there as well, more natural.~~
- Implemented in our `websocket` package in order to migrate back to the nhooyr.io websocket when the 2.0 release is done


go.sum
```diff
@@ -1,6 +1,4 @@
 github.com/Microsoft/go-winio v0.5.2/go.mod h1:WpS1mjBmmwHBEWmogvA2mj8546UReBk4v8QkMxJ6pZY=
-github.com/Microsoft/go-winio v0.6.1 h1:9/kr64B9VUZrLm5YYwbGtUJnMgqWVOdUAXu6Migciow=
```
denopink, line 2: `go mod tidy`

internal/nhooyr.io/websocket/conn.go
```diff
@@ -185,6 +193,9 @@ func (c *Conn) close(err error) {
 }
 
 func (c *Conn) timeoutLoop() {
```
denopink, line 195: `activityProbe` is used to monitor reads and writes activity until the connection is closed.

internal/nhooyr.io/websocket/conn.go
```diff
@@ -141,6 +149,12 @@ func newConn(cfg connConfig) *Conn {
 		c.close(errors.New("connection garbage collected"))
 	})
```
denopink: We used the options patterns to set `inactivityTimeout`, in order to avoid more locks due to the `inactivityTimeout` race condition created in `c.timeoutLoop()`:

```
c.wg.Add(1)
go func() {
	defer c.wg.Done()
	c.timeoutLoop()
}()
```

internal/nhooyr.io/websocket/conn_options.go
```diff
@@ -0,0 +1,30 @@
...
+
+type ConnOptionFunc func(*Conn)
+
+func (f ConnOptionFunc) apply(c *Conn) {
+	f(c)
+}
+
+// InactivityTimeout sets inactivity timeout for the WS connection.
+// Nonpositive InactivityTimeout will default to a 1 minute timeout.
+func InactivityTimeout(d time.Duration) ConnOption {
```
denopink: We used the options patterns to set `inactivityTimeout`, in order to avoid more locks due to the `inactivityTimeout` race condition created in `c.timeoutLoop()`:

```
c.wg.Add(1)
go func() {
	defer c.wg.Done()
	c.timeoutLoop()
}()
```

internal/nhooyr.io/websocket/conn.go
```diff
@@ -210,6 +232,58 @@ func (c *Conn) timeoutLoop() {
 	}
 }
 
+// monitorConnectionActivity monitors whether the connection is active by probing reads & writes. If the
```
denopink:
```suggestion
// monitorConnectionActivity determines whether the connection is active by monitoring for reads & writes activity. If the
```

schmidtw:
Because we want to migrate back to the nhooyr.io websocket when the 2.0 release is done, we should strive to make the changes in just the websocket wrapper in our code.

internal/websocket/ws.go
```diff
@@ -278,13 +292,27 @@ func (ws *Websocket) run(ctx context.Context) {
 						Type: event.PONG,
 					})
 				})
+
+				// Record pong activity.
+				ws.activity <- struct{}{}
```
denopink: remove

internal/websocket/ws.go
```diff
@@ -214,13 +220,18 @@ func (ws *Websocket) Send(ctx context.Context, msg wrp.Message) error {
 		err = ws.conn.Write(ctx, nhws.MessageBinary, wrp.MustEncode(&msg, wrp.Msgpack))
 	}
 	ws.m.Unlock()
+	// Record write activity.
+	ws.activity <- struct{}{}
```
denopink: remove

internal/websocket/ws.go
```diff
@@ -214,13 +220,18 @@ func (ws *Websocket) Send(ctx context.Context, msg wrp.Message) error {
...
 
 	return err
 }
 
 func (ws *Websocket) run(ctx context.Context) {
 	ws.wg.Add(1)
 	defer ws.wg.Done()
+	// Closing ws.activity after we.run() returns insures that conenction activity is not blocked
+	// while the conenction is actively closing.
+	defer close(ws.activity)
```
denopink: make local

internal/websocket/ws.go
```diff
@@ -266,6 +269,8 @@ func (ws *Websocket) run(ctx context.Context) {
 						Type: event.PING,
 					})
 				})
+
+				inactivityTimeout = time.After(ws.inactivityTimeout)
```
denopink, line 273: based on how `typ, reader, err := ws.conn.Reader(ctx)` is implemented, there's no race condition from what I see (`go test --race` comes back clean as well)

internal/websocket/ws.go
```diff
@@ -284,7 +289,21 @@ func (ws *Websocket) run(ctx context.Context) {
 			// Read loop
 			for {
 				var msg wrp.Message
+				ctx, cancel := context.WithTimeout(ctx, ws.inactivityTimeout)
 				typ, reader, err := ws.conn.Reader(ctx)
+
```
schmidtw: I think the call to `cancel()` should be here since that immediately frees up the context.  Otherwise it will linger around until the inactivityTimeout finally comes into effect.

internal/websocket/ws.go
```diff
@@ -293,6 +312,8 @@ func (ws *Websocket) run(ctx context.Context) {
 						err = decoder.Decode(&msg)
 					}
 				}
+				// Cancel ws.conn.Reader()'s context after wrp decoding.
+				cancel()
```
schmidtw, line 321: Move up.  See the comment on line 294.

schmidtw (changes requested):

schmidtw (approved):

internal/websocket/ws.go
```diff
@@ -284,7 +289,21 @@ func (ws *Websocket) run(ctx context.Context) {
 			// Read loop
 			for {
 				var msg wrp.Message
+				ctx, cancel := context.WithTimeout(ctx, ws.inactivityTimeout)
 				typ, reader, err := ws.conn.Reader(ctx)
+				if errors.Is(err, context.DeadlineExceeded) {
+					select {
+					case <-inactivityTimeout:
```
johnabass, line 296: You don't need a channel created via `time.After` here.  Since you've just got a single loop, you can use a (possibly atomic) timestamp that is the expiry time.  Something like:

```
expiry := time.Now().Add( /* whatever the timeout is */ )
for {
    typ, reader, err := ws.conn.Reader(ctx)
    if errors.Is(err, context.DeadlineExceeded) {
        if time.Now().After(expiry) {
             // timeout exceeded
        }

        // new expiry for the next iteration
        expiry = expiry.Add( /* timeout */ )
    }
}
```

If other sources need to update the expiry, then pull it out as a member on the `Websocket` type and treat it as an atomic variable, using the unix timestamp value:

```
type Websocket struct {
    expiry atomic.Int64 // use the new API, it's simpler
}

func (ws *Websocket) Start() {
    // you don't have to use UTC, but it makes debugging easier since no matter where this
    // code is run, the expiry will be in the same TZ
    ws.expiry.Store(time.Now().UTC().Unix())

    // continue the for loop as above, but use ws.expiry.Load and ws.expiry.Store to access the expiry value
}
```

The above allows `Websocket` to have a method that updates the expiry value in a lock-free way that is concurrent-safe.

internal/websocket/ws.go
```diff
@@ -284,7 +289,21 @@ func (ws *Websocket) run(ctx context.Context) {
 			// Read loop
 			for {
 				var msg wrp.Message
+				ctx, cancel := context.WithTimeout(ctx, ws.inactivityTimeout)
 				typ, reader, err := ws.conn.Reader(ctx)
+				if errors.Is(err, context.DeadlineExceeded) {
```
johnabass, line 294: Don't forget that both `context.DeadlineExceeded` and `context.Canceled` can be returned by the `context` package.  `context.Canceled` means that something called the cancel function before the timeout occurred, which would indicate that something external to this type requested shutdown of the `Websocket`.  That's may be a slightly different case, I'm not sure.

internal/websocket/ws.go
```diff
@@ -284,7 +289,21 @@ func (ws *Websocket) run(ctx context.Context) {
 			// Read loop
 			for {
 				var msg wrp.Message
+				ctx, cancel := context.WithTimeout(ctx, ws.inactivityTimeout)
```
johnabass, line 292: You can move the `cancel()` call here to ensure that any previous context got cleaned up prior to reading.  Just do something like:

```
if cancel != nil {
    cancel()
}
```

That's not perfect, as it means that if you exit the loop below, the context is still active.  but you can take care of that by having a parent context and doing a `defer parentCancel()` at the top of this function.

internal/websocket/ws.go
denopink: good point! I think we're ok because our ending select will catch the case when our parent ctx has been cancelled

```go
select {
case <-time.After(next):
case <-ctx.Done():
	return
}
```

internal/websocket/ws.go
denopink: great point! I choose `time.After` since we're already using one in the ending context (trying to be consistent). But that's a different use case where `time.After` makes more sense. I can see how `time.Now().Add()` would fit better here and provide us the opportunity to expose it safely if we ever wanted to :)


```go
select {  
case <-time.After(next):  
case <-ctx.Done():  
    return  
}
```

denopink: @schmidtw do you have a preference?

johnabass: Also, to make this kind of code testable, you don't want to use `time.Now` directly.  Instead, use a `func() time.Time` and you inject into the `Websocket` type.  This is why I created github.com/xmidt-org/chronon.  It makes testing time-related logic easier.

internal/websocket/ws.go
johnabass: Those are still (2) cases though.  A unit test should cover both, and verify that both of those channels being signaled produces the desired behavior in each case.

internal/websocket/ws.go
denopink: I'll implement this in a new pr 🙂

internal/websocket/ws.go
denopink: 💯 just added the test `internal/websocket/ws_test.go: Test_CancelCtx` to cover this case 🙂

PR #142, opened 2024-05-22
remove inline default config

piccione99:
embed config file

please look over for appropriate comments explaining the config


cmd/xmidt-agent/config.go
```diff
@@ -244,11 +246,13 @@ func provideConfig(cli *CLI) (*goschtalt.Config, error) {
 				goschtalt.ValidatorFunc(validate.Validate),
 			),
 		),
+		// Seed the program with the default, built-in configuration
+		goschtalt.AddBuffer("build-in.yaml", defaultConfigFile, goschtalt.AsDefault()),
```
schmidtw: Maybe use the name `!built-in.yaml` so this is lexigraphically very very early.  See asciitable.com for the ordering.

cmd/xmidt-agent/config.go
```diff
@@ -244,11 +246,13 @@ func provideConfig(cli *CLI) (*goschtalt.Config, error) {
...
 
 		// Seed the program with the default, built-in configuration.
 		// Mark this as a default so it is ordered correctly.
-		goschtalt.AddValue("built-in", goschtalt.Root, defaultConfig,
-			goschtalt.AsDefault()),
+		// goschtalt.AddValue("built-in.yaml", goschtalt.Root, defaultConfig,
```
schmidtw: Just remove this.

cmd/xmidt-agent/default-config.yaml
```diff
@@ -0,0 +1,184 @@
...
+    - hw-last-reboot-reason 
+    - webpa-protocol
+    - boot-time
+    - boot-time-retry-wait
+    - webpa-interface-used
+    - interfaces-available
+network_service:
+  allowed_interfaces:
+    eth0:
+      priority: 1
```
schmidtw, line 148: Add some detail about if 1 is higher or lower priority than 9.

schmidtw (changes requested):

cmd/xmidt-agent/config.go
denopink: that's so cool, never knew that!

cmd/xmidt-agent/config.go
```diff
@@ -17,16 +17,17 @@ import (
...
 	"github.com/xmidt-org/retry"
 	"github.com/xmidt-org/sallust"
 	"github.com/xmidt-org/wrp-go/v3"
 	"github.com/xmidt-org/xmidt-agent/internal/configuration"
 	"github.com/xmidt-org/xmidt-agent/internal/net"
-	"go.uber.org/zap/zapcore"
 	"gopkg.in/dealancer/validate.v2"
 )
 
+//go:embed default-config.yaml
```
denopink, line 28: that's neat! 🙂

cmd/xmidt-agent/config.go
```diff
@@ -236,6 +237,7 @@ type NetworkService struct {
 // Collect and process the configuration files and env vars and
 // produce a configuration object.
 func provideConfig(cli *CLI) (*goschtalt.Config, error) {
+
```
denopink, line 240:
```suggestion
```

random newline

cmd/xmidt-agent/default-config.yaml
```diff
@@ -0,0 +1,184 @@
...
+    encode_time:     RFC3339Nano
+    encode_duration: string
+    encode_caller:   short
+  rotation: 
+    max_size:    1 #  1MB max/file
+    max_age:     30 # 30 days max
+    MaxBackups: 10 # max 10 files
+operational_state:
+  last_reboot_reason: sleepy
+  boot_time: "2024-02-28T01:04:27Z"
```
denopink: should we have a default for `boot_time`? Maybe xmidt-agent can use `time.Now()` when `boot_time` is empty

cmd/xmidt-agent/default-config.yaml
```diff
@@ -0,0 +1,184 @@
...
+  temporary: "~/local-rdk-testing/temporary"
+  durable: "~/local-rdk-testing/durable"
+mock_tr_181:
+  enabled: true
+  file_path: "mock_tr181.json"
+  service_name: "mock_config"
+xmidt_agent_crud:
+  service_name: xmidt_agent
+qos: 
+  max_queue_bytes:  1000000 # 1 * 1024 * 1024 // 1MB max/queue,
```
denopink:
```suggestion
  max_queue_bytes:  1048576 # 1 * 1024 * 1024 // 1MB max/queue,
```

cmd/xmidt-agent/default-config.yaml
```diff
@@ -0,0 +1,184 @@
...
+  rotation: 
+    max_size:    1 #  1MB max/file
+    max_age:     30 # 30 days max
+    MaxBackups: 10 # max 10 files
+operational_state:
+  last_reboot_reason: sleepy
+  boot_time: "2024-02-28T01:04:27Z"
+storage:
+  temporary: "~/local-rdk-testing/temporary"
+  durable: "~/local-rdk-testing/durable"
```
denopink: @schmidtw do we want storage by default?

cmd/xmidt-agent/default-config.yaml
```diff
@@ -0,0 +1,184 @@
+# SPDX-FileCopyrightText: 2023 Comcast Cable Communications Management, LLC
```
denopink, line 1: formatting looks a little off (like spacing, i.e. `ping_write_timeout`), tried running a yaml linter? (if you haven't already 🙂 )

denopink (changes requested):
I made a few suggestions, looking good so far 🙂

cmd/xmidt-agent/default-config.yaml
schmidtw: The problem with a boot time that is wrong but not clearly wrong is that it's really hard to disambiguate later.  If we leave it empty and then use the value that `time.Time{}` gives us then we know the value was not set.  I think this value is like `0001-01-01T00:00:00Z`.

cmd/xmidt-agent/default-config.yaml
schmidtw: No.

cmd/xmidt-agent/default-config.yaml
piccione99: do you know of one that doesn't remove comments?  I've tried a few and they all wipe out comments

cmd/xmidt-agent/default-config.yaml
piccione99: see updated date for boot_time

cmd/xmidt-agent/default-config.yaml
denopink: yeah... I just realized I also disabled mine because of that. I guess I got tired of it as well and manually formatted it 😭 

There has to be one that doesn't delete comments.

denopink (approved):
lgtm! 🍻

cmd/xmidt-agent/default-config.yaml
piccione99: there is some python one but it looked multi step and seemed like a pia

schmidtw: I'll add a linter action to the repo for this file.

schmidtw (approved):

PR #147, opened 2024-05-29
fix dockerfile dependency location

piccione99:
moved a copy of mock_tr181.json into the .release/docker directory so the release can build


denopink (approved):
lgtm! 🍻

PR #148, opened 2024-05-29
feat: add a timestamp component to qos in order to prioritize older/newer messages

denopink:
- 100% test cover for `internal/wrphandlers/qos`
- added `PrioritizeOldest` options for qos
- if `PrioritizeOldest` is false, then qos will favor the newest message during a qos value tie breaker (vice versa)
~~- Updated `priorityQueue.trim()` to remove messages with the lowest qos value (taking `PrioritizeOldest` into account as well)~~


internal/wrphandlers/qos/priority_queue.go
```diff
@@ -51,34 +64,69 @@ func (pq *priorityQueue) Enqueue(msg wrp.Message) error {
 	return nil
 }
 
+// trim removes messages with the lowest QualityOfService (taking `prioritizeOldest` into account)
+// until the queue no longer violates `maxQueueSize“.
 func (pq *priorityQueue) trim() {
```
denopink: Introduced this since we described a desire to keep messages with higher qos value taking the age of the message into account as well

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -51,34 +64,69 @@ func (pq *priorityQueue) Enqueue(msg wrp.Message) error {
...
+		// Re-prioritize messages with the highest QualityOfService.
+		pq.prioritizeLowestQOS = false
+		// heap.Init() is required since the prioritization was switch to messages with the highest QualityOfService.
+		// The complexity of heap.Init() is O(n) where n = h.Len().
+		heap.Init(pq)
+	}()
+
+	// heap.Init() is required since the prioritization was switch to messages with the lowest QualityOfService.
+	// The complexity of heap.Init() is O(n) where n = h.Len().
+	heap.Init(pq)
```
denopink: If we end up needing to trim often, then the `O(n)` cost  of `heap.Init()` may start to sting...

I wanted to point that out in case we can't afford the `O(n)` cost  of `heap.Init()` and we would rather trim() random messages instead.

cmd/xmidt-agent/config.go
```diff
@@ -62,6 +62,9 @@ type QOS struct {
 	MaxQueueBytes int64
 	// MaxMessageBytes is the largest allowable wrp message payload.
 	MaxMessageBytes int
+	// PrioritizeOldest determines whether to prioritize the oldest message during a QualityOfService tie breaker,
+	// with the default being to prioritize the newest messages.
+	PrioritizeOldest bool
```
schmidtw: I think `Priorize string` where the string values are `newest`, `oldest` to start with is more clear what is happening.  It also allows expansion later if we need different forms of prioritization, etc.

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -17,7 +18,14 @@ var ErrMaxMessageBytes = errors.New("wrp message payload exceeds maxMessageBytes
 // https://xmidt.io/docs/wrp/basics/#qos-description-qos
 type priorityQueue struct {
 	// queue for wrp messages, ingested by serviceQOS
-	queue []wrp.Message
+	queue []item
+	// prioritizeOldest determines whether to prioritize the oldest message during a QualityOfService tie breaker,
+	// with the default being to prioritize the newest messages.
+	prioritizeOldest bool
+	// prioritizeLowestQOS determines whether to prioritize messages with the lowest QualityOfService,
```
schmidtw: Can the `prioritzeLowestQOS` optimization be done in a separate PR?

internal/wrphandlers/qos/options.go
```diff
@@ -45,3 +45,14 @@ func MaxMessageBytes(s int) Option {
 			return nil
 		})
 }
+
+// PrioritizeOldest determines whether to prioritize the oldest message during a QualityOfService tie breaker,
+// with the default being to prioritize the newest messages.
+func PrioritizeOldest(b bool) Option {
+	return optionFunc(
+		func(h *Handler) error {
+			h.prioritizeOldest = b
```
schmidtw: Instead of setting a boolean here, use a function that does the comparison.  That will let you remove a few conditional `if` statements elsewhere.

schmidtw (changes requested):

cmd/xmidt-agent/config.go
denopink: That's a great point!

internal/wrphandlers/qos/priority_queue.go
denopink: I agree, that'll make this pr more easily digestible and focused.

I'll be more conscious of that in future prs 🙂

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -69,16 +78,42 @@ func (pq *priorityQueue) drop() {
 func (pq *priorityQueue) Len() int { return len(pq.queue) }
 
 func (pq *priorityQueue) Less(i, j int) bool {
-	return pq.queue[i].QualityOfService > pq.queue[j].QualityOfService
+	iQOS := pq.queue[i].msg.QualityOfService
+	jQOS := pq.queue[j].msg.QualityOfService
+	iTimestamp := pq.queue[i].timestamp
+	jTimestamp := pq.queue[j].timestamp
+
```
schmidtw: Reorder this function so the extra work is only done if needed.

```golang
    if iQOS != jQOS {
        return iQOS > jQOS
    }
```

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -69,16 +78,42 @@ func (pq *priorityQueue) drop() {
...
+	// Determine what will be used as a priority tie breaker.
+	var tieBreaker bool
+	switch pt := pq.priority; pt {
```
schmidtw: Instead of a switch here, make a function that is set by the options & validated when the PQ is created.  That saves a check here each time & the need to handle a panic scenario.

schmidtw (changes requested):

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -69,16 +78,42 @@ func (pq *priorityQueue) drop() {
...
+	case NewestType:
+		// Prioritize the newest messages.
+		tieBreaker = iTimestamp.After(jTimestamp)
+	case OldestType:
+		// Prioritize the oldest messages.
+		tieBreaker = iTimestamp.Before(jTimestamp)
+	default:
+		panic(errors.Join(ErrPriorityTypeInvalid, fmt.Errorf("PriorityType error: '%s' does not match any valid options: %s",
+			pt, pt.getKeys())))
```
```suggestion
		panic(errors.Join(ErrPriorityTypeInvalid, fmt.Errorf("PriorityType error: '%s' does not match any valid options",
			pt)))
```

internal/wrphandlers/qos/priority_queue.go
denopink: Oh ok, it didn't click the first time but now I get it 🙂

schmidtw (approved):

PR #152, opened 2024-06-04
feat: optimize qos queue trimming with `heap.Init()` (`O(n)`)

denopink:
- 100% test cover for `internal/wrphandlers/qos`
- optimize `priorityQueue.trim()` to remove messages with the lowest qos value while using `trimTieBreaker` breaks any QualityOfService ties during queue trimming
- the compute cost of optimizing `priorityQueue.trim()` will be O(n)

More details on the compute cost of optimizing trim:

When we trim and we're looking to trim the least prioritized messages, then (right before removing messages) we need to call heap.Init() since the prioritization was changed. https://pkg.go.dev/container/heap#Init

The complexity of heap.Init() is O(n). Every time we need to trim, heap.Init() will be called twice (before and after trimming).

So the tax for trimming is: O(n) `[two heap.Init()  calls]` + O(n*log(n)) `[X calls to sort and remove messages from the queue]`

### DIFF
--- a/internal/wrphandlers/qos/options.go
+++ b/internal/wrphandlers/qos/options.go
@@ -51,21 +51,31 @@ func MaxMessageBytes(s int) Option {
 // with the default being to prioritize the newest messages.
 func Priority(p PriorityType) Option {
 	return optionFunc(
-		func(h *Handler) error {
-			// Determine what will be used as a QualityOfService tie breaker.
-			switch p {
-			case NewestType:
-				// Prioritize the newest messages.
-				h.tieBreaker = PriorityNewestMsg
-			case OldestType:
-				// Prioritize the oldest messages.
-				h.tieBreaker = PriorityOldestMsg
-			default:
-				return errors.Join(fmt.Errorf("%w: %s", ErrPriorityTypeInvalid, h.priority), ErrMisconfiguredQOS)
-			}
-
+		func(h *Handler) (err error) {
+			h.tieBreaker, h.trimTieBreaker, err = priority(p)
 			h.priority = p
 
-			return nil
+			return err
 		})
+}
+
+// priority determines which tie breakers are used during normal enqueueing and queue trimming.
+func priority(p PriorityType) (enqueueTieBreaker tieBreaker, trimTieBreaker tieBreaker, err error) {
+	// Determine what will be used as a QualityOfService tie breaker during normal enqueueing and queue trimming.
+	switch p {
+	case NewestType:
+		// Prioritize the newest messages.
+		enqueueTieBreaker = PriorityNewestMsg
+		// Remove the oldest messages during trimming.
+		trimTieBreaker = PriorityOldestMsg
+	case OldestType:
+		// Prioritize the oldest messages.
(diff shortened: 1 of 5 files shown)


internal/wrphandlers/qos/options.go
```diff
@@ -51,21 +51,31 @@ func MaxMessageBytes(s int) Option {
...
+			h.tieBreaker, h.trimTieBreaker, err = priority(p)
 			h.priority = p
 
-			return nil
+			return err
 		})
 }
+
+// priority determines which tie breakers are used during normal enqueueing and queue trimming.
+func priority(p PriorityType) (enqueueTieBreaker tieBreaker, trimTieBreaker tieBreaker, err error) {
```
denopink, line 63: Making this its own function helps with testing.

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -61,19 +67,33 @@ func (pq *priorityQueue) Enqueue(msg wrp.Message) error {
 	return nil
 }
 
+// trim removes messages with the lowest QualityOfService (taking `prioritizeOldest` into account)
+// until the queue no longer violates `maxQueueSize“.
 func (pq *priorityQueue) trim() {
```
denopink, line 71: when we trim and we're looking to trim the least prioritized messages, then (right before removing messages) we need to call heap.Init() since the prioritization was changed. https://pkg.go.dev/container/heap#Init

The complexity of heap.Init() is O(n). Every time we need to trim, heap.Init() will be called twice (before and after trimming).

So the tax for trimming is: O(n) `[two heap.Init()  calls]` + O(n*log(n)) `[X calls to sort and remove messages from the queue]`

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -21,6 +21,12 @@ type priorityQueue struct {
 	queue []item
 	// tieBreaker breaks any QualityOfService ties.
 	tieBreaker tieBreaker
+	// queueInTrimming indicates the queue is actively being trimmed, where messages with the lowest
+	// QualityOfService are prioritize and removed.
```
```suggestion
	// QualityOfService are removed.
```

internal/wrphandlers/qos/priority_queue.go
denopink: I think this is the simplest way we can achieve this while minimizing code changes.

I think the cost of this new trimming code will be small overall because I think the situation where we are trimming often is a very unlikely edge case, it’ll require the xmidt-agent to
- consistently or completely be unable to send messages to xmidt (e.g outage or connectivity issues)
- while also receiving a steady flow of new messages to a already full queued

One way i think we can mitigate this edge case of constantly trimming may be (during a single pass of trimming) to remove stale messages (if any) and if there aren't any stale messages, then the lowest qos message can be removed.

that way we're potentially removing bulks of stale messages rather than removing 1 message at a time during a single pass of trimming

denopink: but if we want to optimize trimming further in the future, I think we can achieve n*log(n) trimming (I have a working version locally but it introduces a bit of code)

PR #155, opened 2024-06-12
base64 encode metadata header

piccione99:


internal/metadata/metadata.go
```diff
@@ -11,6 +11,7 @@ import (
 
 	"github.com/xmidt-org/wrp-go/v3"
 	"github.com/xmidt-org/xmidt-agent/internal/net"
+	b64 "encoding/base64"
```
schmidtw: unless `base64` is conflicting with something I'd use that, otherwise this looks good.

schmidtw (approved):

PR #160, opened 2024-06-18
feat: add support for mock181's SET to update parameter states

denopink:
- mock181's SET didn't update parameter state
- add support for mock181's  SET to update parameter states
- return the updated parameters as the response body for mock181's  SET


internal/wrphandlers/mocktr181/handler.go
```diff
@@ -181,20 +184,34 @@ func (h Handler) get(names []string) (int64, []byte, error) {
...
 					mockParameter.Value = parameter.Value
 					mockParameter.DataType = parameter.DataType
 					mockParameter.Attributes = parameter.Attributes
+
+					result.Parameters = append(result.Parameters, Parameter{
+						Name:       mockParameter.Name,
+						Value:      mockParameter.Value,
+						DataType:   mockParameter.DataType,
+						Attributes: mockParameter.Attributes,
+					})
```
denopink: @ilawjr asked for mock181 SET's response payload to contain the update parameters

internal/wrphandlers/mocktr181/handler.go
```diff
@@ -181,20 +184,34 @@ func (h Handler) get(names []string) (int64, []byte, error) {
 	return statusCode, payload, nil
 }
 
-func (h Handler) set(parameters []Parameter) int64 {
+func (h Handler) set(parameters []Parameter) (int64, []byte, error) {
+	result := Tr181Payload{}
 	for _, parameter := range parameters {
-		for _, mockParameter := range h.parameters {
+		for i := range h.parameters {
+			mockParameter := &h.parameters[i]
```
denopink, line 204: I think we meant to use `mockParameter` as a reference, otherwise the value update will not stick (i.e. mock GET will get the default value since it'll never update)

internal/wrphandlers/mocktr181/handler.go
```diff
@@ -181,20 +184,34 @@ func (h Handler) get(names []string) (int64, []byte, error) {
...
 			if strings.HasPrefix(mockParameter.Name, parameter.Name) {
 				if mockParameter.Access == "rw" {
```
schmidtw: If `mockParameter.Access` == `"r"` or something else, a set will silently fail.

internal/wrphandlers/mocktr181/handler.go
```diff
@@ -181,20 +185,42 @@ func (h Handler) get(names []string) (int64, []byte, error) {
...
+						Value:      mockParameter.Value,
+						DataType:   mockParameter.DataType,
+						Attributes: mockParameter.Attributes,
+					})
+				default:
+					result.Parameters = append(result.Parameters, Parameter{
+						Name:    mockParameter.Name,
+						Message: "Parameter is not writable",
+					})
+					statusCode = 520
```
denopink: Example of what we're expecting:

```
{
  "parameters": [
    {
      "name": "Device.Bridging.Bridge.8.Status",
      "message": "Parameter is not writable"
    }
  ],
  "statusCode": 520
}
```

internal/wrphandlers/mocktr181/handler.go
```diff
@@ -158,13 +162,24 @@ func (h Handler) get(names []string) (int64, []byte, error) {
...
+
+			switch mockParameter.Access {
+			case "r", "rw", "wr":
 				result.Parameters = append(result.Parameters, Parameter{
 					Name:       mockParameter.Name,
 					Value:      mockParameter.Value,
 					DataType:   mockParameter.DataType,
 					Attributes: mockParameter.Attributes,
 				})
+			default:
```
denopink, line 177: For a non-readable parameter:
```
{
                "name": "Device.Bridging.Bridge.8.VLANNumberOfEntries",
                "value":        "1",
                "access":       "w",
                "type": 2
 }
```

This is the expected response:
```
{
  "message": "Invalid parameter name",
  "statusCode": 520
}
```

schmidtw (approved):

PR #162, opened 2024-06-18
patch: update auth wrphandler to be webpa compliant

denopink:

### DIFF
--- a/internal/wrphandlers/auth/handler.go
+++ b/internal/wrphandlers/auth/handler.go
@@ -87,11 +87,11 @@ func (h Handler) HandleWrp(msg wrp.Message) error {
 	response := msg
 	response.Destination = msg.Source
 	response.Source = h.source
-	response.ContentType = "text/plain"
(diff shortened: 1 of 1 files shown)


internal/wrphandlers/auth/handler.go
```diff
@@ -87,11 +87,11 @@ func (h Handler) HandleWrp(msg wrp.Message) error {
 	response := msg
 	response.Destination = msg.Source
 	response.Source = h.source
-	response.ContentType = "text/plain"
-	response.Payload = []byte(fmt.Sprintf("Partner(s) '%s' not allowed.  Allowed: '%s'", got, want))
+	response.ContentType = "application/json"
 
 	code := int64(statusCode)
 	response.Status = &code
+	response.Payload = []byte(fmt.Sprintf(`{statusCode: %d, message:"Partner(s) '%s' not allowed.  Allowed: '%s'"}`, code, got, want))
```
denopink, line 94: the device has to put the status code in its payload as a json https://github.com/xmidt-org/tr1d1um/blob/a8744affcbfa79ef19d018585a93c31843e0e7b2/translation/transport.go#L214-L216

schmidtw (approved):

PR #163, opened 2024-06-18
patch: update missing wrphandler to be webpa compliant

denopink:

### DIFF
--- a/internal/wrphandlers/missing/handler.go
+++ b/internal/wrphandlers/missing/handler.go
@@ -70,10 +70,11 @@ func (h Handler) HandleWrp(msg wrp.Message) error {
 	response := msg
 	response.Destination = msg.Source
 	response.Source = h.source
-	response.Payload = nil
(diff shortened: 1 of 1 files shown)


internal/wrphandlers/missing/handler.go
```diff
@@ -70,10 +70,11 @@ func (h Handler) HandleWrp(msg wrp.Message) error {
 	response := msg
 	response.Destination = msg.Source
 	response.Source = h.source
-	response.Payload = nil
+	response.ContentType = "application/json"
 
 	code := int64(statusCode)
 	response.Status = &code
+	response.Payload = []byte(fmt.Sprintf("{statusCode: %d}", code))
```
denopink, line 77: the device has to put the status code in its payload as a json https://github.com/xmidt-org/tr1d1um/blob/a8744affcbfa79ef19d018585a93c31843e0e7b2/translation/transport.go#L214-L216

PR #164, opened 2024-06-18
patch: update xmidt_agent_crud wrphandler to be webpa compliant

denopink:
- xmidt_agent_crud will now response back when an error occurs

### DIFF
--- a/internal/wrphandlers/xmidt_agent_crud/handler.go
+++ b/internal/wrphandlers/xmidt_agent_crud/handler.go
@@ -5,6 +5,7 @@ package xmidt_agent_crud
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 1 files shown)


internal/wrphandlers/xmidt_agent_crud/handler.go
```diff
@@ -36,31 +37,35 @@ func New(egress wrpkit.Handler, source string, logLevel loglevel.LogLevel) (*Han
...
 	payload := make(map[string]string)
 
 	err := json.Unmarshal(msg.Payload, &payload)
-	if err != nil { // do we still want to return a response here?
-		return err // this would be sent to the eventor error handler
+	if err != nil {
+		statusCode := int64(http.StatusInternalServerError)
+		response.Status = &statusCode
+		response.Payload = []byte(fmt.Sprintf(`{statusCode: %d, message: "%s"}`, statusCode, err.Error()))
+		return h.egress.HandleWrp(response)
```
denopink, line 51: I think we should response back with the error otherwise it'll make trouble shooting harder than it should be

internal/wrphandlers/xmidt_agent_crud/handler.go
```diff
@@ -36,31 +37,35 @@ func New(egress wrpkit.Handler, source string, logLevel loglevel.LogLevel) (*Han
...
+	response.ContentType = "application/json"
 	payload := make(map[string]string)
 
 	err := json.Unmarshal(msg.Payload, &payload)
-	if err != nil { // do we still want to return a response here?
-		return err // this would be sent to the eventor error handler
+	if err != nil {
+		statusCode := int64(http.StatusInternalServerError)
+		response.Status = &statusCode
+		response.Payload = []byte(fmt.Sprintf(`{statusCode: %d, message: "%s"}`, statusCode, err.Error()))
```
denopink, line 50: the device has to put the status code in its payload as a json https://github.com/xmidt-org/tr1d1um/blob/a8744affcbfa79ef19d018585a93c31843e0e7b2/translation/transport.go#L214-L216

schmidtw (approved):

PR #182, opened 2024-07-08
patch: add missing mocktr181's response fields

denopink:
- add the missing `statusCode` field to mocktr181's response payload
- add the expected values for the response fields `Command` and `Names`
- add missing `Success` message for successful `gets` and `patches`
- add missing `Count` field to mocktr181's response payload

### DIFF
--- a/internal/wrphandlers/mocktr181/handler.go
+++ b/internal/wrphandlers/mocktr181/handler.go
@@ -167,7 +167,6 @@ func (h Handler) get(tr181 *Tr181Payload) (int64, []byte, error) {
 
 	for _, name := range tr181.Names {
 		for _, mockParameter := range h.parameters {
-			result.StatusCode = http.StatusOK
 			if !strings.HasPrefix(mockParameter.Name, name) {
 				continue
 			}
@@ -182,6 +181,7 @@ func (h Handler) get(tr181 *Tr181Payload) (int64, []byte, error) {
 					Message:    "Success",
 					Count:      1,
 				})
+				result.StatusCode = http.StatusOK
 			default:
 				result.Parameters = append(result.Parameters, Parameter{
 					Message: fmt.Sprintf("Invalid parameter name: %s", mockParameter.Name),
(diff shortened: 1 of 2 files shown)


internal/wrphandlers/mocktr181/handler.go
```diff
@@ -178,26 +180,27 @@ func (h Handler) get(names []string) (int64, []byte, error) {
...
+
 	payload, err := json.Marshal(result)
 	if err != nil {
 		return http.StatusInternalServerError, payload, errors.Join(ErrInvalidResponsePayload, err)
 	}
 
-	if len(result.Parameters) == 0 {
-		statusCode = int64(520)
-	}
-
```
denopink, line 194: this is out of order, should be set before payload marshaling

internal/wrphandlers/mocktr181/handler.go
```diff
@@ -173,32 +178,37 @@ func (h Handler) get(names []string) (int64, []byte, error) {
 					Value:      mockParameter.Value,
 					DataType:   mockParameter.DataType,
 					Attributes: mockParameter.Attributes,
+					Message:    "Success",
+					Count:      1,
```
denopink, line 182: note: non-wildcard `get`s will always have a Count of 1 (based on `parodus2ccsp` and `wdmp-c`)

internal/wrphandlers/mocktr181/handler.go
```diff
@@ -173,32 +178,37 @@ func (h Handler) get(names []string) (int64, []byte, error) {
...
 				result.Parameters = append(result.Parameters, Parameter{
 					Message: fmt.Sprintf("Invalid parameter name: %s", mockParameter.Name),
 				})
-				statusCode = 520
+				result.StatusCode = 520
 			}
 		}
 	}
 
+	if len(result.Parameters) == 0 {
```
schmidtw: If you default the `StatusCode` to `520` and only set it when appending to the parameters on line 171, then no conditional check is needed here.

schmidtw (changes requested):

schmidtw (approved):

PR #192, opened 2024-07-15
patch: update mock181's `set` to check for non-writable params before updating params and for non-existing parameters

denopink:
During mock181's `set`:
- check for any non-writable params
- if any param is not writable, do not update any param
- add case when request param is not found

### DIFF
--- a/internal/wrphandlers/mocktr181/handler.go
+++ b/internal/wrphandlers/mocktr181/handler.go
@@ -205,6 +205,12 @@ func (h Handler) set(tr181 *Tr181Payload) (int64, []byte, error) {
 		Names:      tr181.Names,
 		StatusCode: http.StatusAccepted,
 	}
+
+	var (
+		writableParams []*MockParameter
+		failedParams   []Parameter
+	)
+	// Check for any parameters that are not writable.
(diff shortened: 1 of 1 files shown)


denopink (commented):
FYI, I'll have to rewrite the tests for mock181 so that it verifies failure cases like rollback was successful during a `set` failure.

I'll throw that in another pr

internal/wrphandlers/mocktr181/handler.go
```diff
@@ -205,32 +205,70 @@ func (h Handler) set(tr181 *Tr181Payload) (int64, []byte, error) {
...
-					Name:    mockParameter.Name,
-					Message: "Parameter is not writable",
-				})
-				result.StatusCode = 520
+			// Check whether mockParameter is writable.
+			if strings.Contains(mockParameter.Access, "w") {
+				found = true
+				// Add mockParameter to the list of parameters to be updated.
+				writableParams = append(writableParams, mockParameter)
+				continue
```
denopink, line 227:
```suggestion
				// Move onto the next requested parameter.
				break
```

@schmidtw can we assume there aren't any duplicates?

schmidtw (approved):

PR #194, opened 2024-07-15
patch: update mock181's get, check for non-readable or non-existing parameters

denopink:
During mock181's get:
- check for any non-readable params
- if any param is not readable, do not return any parameters that succeeded
- add case when request param is not found
- update wildcard case

### DIFF
--- a/internal/wrphandlers/mocktr181/handler.go
+++ b/internal/wrphandlers/mocktr181/handler.go
@@ -190,7 +190,7 @@ func (h Handler) get(tr181 *Tr181Payload) (int64, []byte, error) {
 				continue
 			}
 
-			// If requested parameter is a wild card and mockParameter is not readable,
+			// If the requested parameter is a wild card and is not readable,
 			// then continue and don't count it as a failure.
 			if name[len(name)-1] == '.' {
 				continue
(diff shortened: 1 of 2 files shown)


internal/wrphandlers/mocktr181/handler.go
```diff
@@ -162,35 +162,54 @@ func (h Handler) get(tr181 *Tr181Payload) (int64, []byte, error) {
...
-				})
-				result.StatusCode = 520
+				continue
 			}
+
+			// mockParameter is not readable.
+			failedNames = append(failedNames, mockParameter.Name)
+		}
+
+		if !found {
```
denopink, line 204: added case when a requested parameter is not found

internal/wrphandlers/mocktr181/handler.go
```diff
@@ -162,33 +162,58 @@ func (h Handler) get(tr181 *Tr181Payload) (int64, []byte, error) {
...
-				result.Parameters = append(result.Parameters, Parameter{
-					Message: fmt.Sprintf("Invalid parameter name: %s", mockParameter.Name),
-				})
-				result.StatusCode = 520
+				continue
 			}
+
+			// If requested parameter is a wild card and mockParameter is not readable,
+			// then continue and don't count it as a failure.
+			if name[len(name)-1] == '.' {
```
denopink, line 196: update wildcard case

schmidtw (approved):

PR #203, opened 2024-07-31
feat: count ping as websocket activity

denopink:
- count ping as websocket activity
- lack of activity will trigger inactivityTimeout, causing the websocket connection to close and reconnect

### DIFF
--- a/internal/websocket/ws.go
+++ b/internal/websocket/ws.go
@@ -228,7 +228,6 @@ func (ws *Websocket) run(ctx context.Context) {
 	mode := ws.nextMode(ipv4)
 
 	policy := ws.retryPolicyFactory.NewPolicy(ctx)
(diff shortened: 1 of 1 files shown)


internal/websocket/ws.go
```diff
@@ -289,22 +291,32 @@ func (ws *Websocket) run(ctx context.Context) {
 			// Read loop
 			for {
 				var msg wrp.Message
-				ctx, cancel := context.WithTimeout(ctx, ws.inactivityTimeout)
-				typ, reader, err := ws.conn.Reader(ctx)
```
denopink, line 293: we want to include the activity from pings. that requires us to manually trigger a timeout.

Otherwise, the device will reconnect in intervals of `ws.inactivityTimeout`

johnabass (approved):
that looks good!

PR #205, opened 2024-07-31
patch: mocktr181 - adding missing `statusCode` and update content type

denopink:
- keep contenttype consistent with other wrp.handlers
- include a `statusCode` for bad mocktr181 cmds


johnabass (approved):
good consistency

PR #214, opened 2024-08-06
populate interface-used with default route

piccione99:


piccione99:
we definitely don't want to do this, just use the input interface

PR #215, opened 2024-08-06
patch: simplify mocktr181 HandleWrp and handle empty payload case

denopink:
- support empty payload case (520 status code)
- simplify mocktr181 HandleWrp

### DIFF
--- a/internal/wrphandlers/mocktr181/handler.go
+++ b/internal/wrphandlers/mocktr181/handler.go
@@ -115,35 +115,9 @@ func (h Handler) Enabled() bool {
 
 // HandleWrp is called to process a tr181 command
 func (h Handler) HandleWrp(msg wrp.Message) error {
-	payload := new(Tr181Payload)
-
-	err := json.Unmarshal(msg.Payload, &payload)
+	statusCode, payloadResponse, err := h.proccessCommand(msg.Payload)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/wrphandlers/mocktr181/handler.go
```diff
@@ -115,53 +115,52 @@ func (h Handler) Enabled() bool {
...
 }
 
+func (h Handler) proccessCommand(wrpPayload []byte) (int64, []byte, error) {
+	var (
+		err             error
+		payloadResponse []byte
+		statusCode      = int64(520)
+	)
+
+	if len(wrpPayload) == 0 {
```
denopink, line 143: support empty payload case (520 status code)

internal/wrphandlers/mocktr181/handler.go
```diff
@@ -115,53 +115,52 @@ func (h Handler) Enabled() bool {
...
+		return statusCode, []byte(fmt.Sprintf(`{"message": ""Invalid Input Command"", "statusCode": %d}`, statusCode)), nil
```
piccione99, line 144: can we say empty so we know the payload is empty?

denopink: I think we would but that's the expected behavior of the old c simulator.

we're trying to match the old c simulator as close as possible

PR #232, opened 2024-09-13
feat: update rdr codes for dropped message

### LINKED ISSUE #208: how should the xmidt-agent respond when messages are dropped either due to their size or due to queue trimming
There appears to be three cases.

How should the xmidt-agent respond when:
1. a request produces a respond that is too large and it's dropped
    1. Should it respond with a `413 Content Too Large`?
2. an event that is too large and dropped
3. a respond or event is trimmed

denopink:
For dropped messages, dispose of the payload and update the rdr code to the following:
- rdr value of 4 for payloads too large
- rdr value of 102 for lower priority messages

### DIFF
--- a/internal/wrphandlers/qos/priority_queue.go
+++ b/internal/wrphandlers/qos/priority_queue.go
@@ -6,6 +6,7 @@ package qos
 import (
 	"container/heap"
 	"errors"
+	"fmt"
 	"slices"
 	"time"
 
@@ -14,7 +15,7 @@ import (
 
 var ErrMaxMessageBytes = errors.New("wrp message payload exceeds maxMessageBytes")
 
-var (
+const (
 	// https://xmidt.io/docs/wrp/basics/#request-delivery-response-rdr-codes
 	messageIsTooLarge                int64 = 4
 	higherPriorityMessageTookTheSpot int64 = 102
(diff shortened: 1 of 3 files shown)


internal/wrphandlers/qos/priority_queue.go
```diff
@@ -56,40 +61,44 @@ type item struct {
...
-func (pq *priorityQueue) Dequeue() (wrp.Message, bool) {
-	var (
-		msg wrp.Message
-		ok  bool
-	)
-	for pq.Len() != 0 {
-		itm := heap.Pop(pq).(item)
-		// itm.discard will be true if `itm` has been marked to be discarded,
-		// i.e. trimmed by `pq.trim()'.
-		if itm.discard {
```
denopink, line 69: Previously we were discarding the message without sending any response back to xmidt

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -15,6 +14,12 @@ import (
 
 var ErrMaxMessageBytes = errors.New("wrp message payload exceeds maxMessageBytes")
 
+var (
+	// https://xmidt.io/docs/wrp/basics/#request-delivery-response-rdr-codes
+	messageIsTooLarge                int64 = 4
```
schmidtw, line 20: Instead of making these globals so they can be returned as a pointer, I'd make them a const an where used create a local variable & return the pointer.

A mistake where an int is written to will change the value forever is my concern.

internal/wrphandlers/qos/priority_queue.go
```diff
@@ -56,40 +61,44 @@ type item struct {
...
 
 // Enqueue queues the given message.
-func (pq *priorityQueue) Enqueue(msg wrp.Message) error {
+func (pq *priorityQueue) Enqueue(msg wrp.Message) {
 	// Check whether msg violates maxMessageBytes.
 	// The zero value of `pq.maxMessageBytes` will disable individual message size validation.
 	if pq.maxMessageBytes != 0 && len(msg.Payload) > pq.maxMessageBytes {
-		return fmt.Errorf("%w: %v", ErrMaxMessageBytes, pq.maxMessageBytes)
+		msg.Payload = nil
+		msg.RequestDeliveryResponse = &messageIsTooLarge
```
schmidtw: Keep the error message so we can see this happened.

schmidtw (approved):

PR #290, opened 2025-03-27
Quic Demo 

piccione99:
Further refactoring to take place after this commit.  Quic and Websocket should both be able to run interchangeably, depending on server. 

2. You can only run quic or websocket, you can't run both.  I vaguely remember talking about being able to run both, but the current structure of the code makes that ugly.   
3.  Quic and Websocket are abstracted by the interfaces in the cloud package. 
4. For now, if the config indicates we are using a redirect server, quic will use http to connect to the redirect server and retrieve the url.  It will then create the quic connection to the final target address. This is because I haven't discovered yet how to retrieve the new quic connection on the client after the redirect takes place.  
5. event interfaces are moved from websocket to their own package since quic needs them too. 
6. Quic is not doing anything to prefer ipv6, go already does this with "Happy Eyeballs" which is enabled


cmd/xmidt-agent/default-config.yaml
```diff
@@ -122,7 +149,7 @@ storage:
   # temporary: "~/local-rdk-testing/temporary"
   # durable: "~/local-rdk-testing/durable"
 mock_tr_181:
-  enabled: false
+  enabled: true
```
piccione99: I will fix the defaults in the final PR

cmd/xmidt-agent/default-config.yaml
```diff
@@ -63,6 +64,32 @@ websocket:
       tls_handshake_timeout:   10s
       expect_continue_timeout: 1s
   max_message_bytes: 262144 # 256 * 1024
+quic:
+  disable: false
```
piccione99: I will fix the defaults in the final PR

cmd/xmidt-agent/default-config.yaml
```diff
@@ -48,6 +48,7 @@ xmidt_service:
 #     -----END PUBLIC KEY-----
 #   timeout: 10s
 websocket:
+  disable: true
```
piccione99: I will fix the defaults in the final PR

cmd/xmidt-agent/config.go
```diff
@@ -119,6 +121,34 @@ type Websocket struct {
...
+	// SendTimeout is the send timeout for the WS connection.
+	SendTimeout time.Duration
+	// AdditionalHeaders are any additional headers for the WS connection.
+	AdditionalHeaders http.Header
+	// FetchURLTimeout is the timeout for the fetching the Quic url. If this is not set, the default is 30 seconds.
+	FetchURLTimeout time.Duration
+	// The client to used to connect to the redirect server
+	HttpClient arrangehttp.ClientConfig
+	// Config for quic connection
+	QuicClient myquic.Http3ClientConfig
```
schmidtw, line 140: This is probably going to get us into trouble.  The `tls.Config` inside this needs an arrangehttp style of help.

cmd/xmidt-agent/config.go
```diff
@@ -119,6 +121,34 @@ type Websocket struct {
...
+	// BackUpURL is the back up XMiDT service endpoint in case `XmidtCredentials.URL` fails.
+	BackUpURL string
+	// SendTimeout is the send timeout for the WS connection.
+	SendTimeout time.Duration
+	// AdditionalHeaders are any additional headers for the WS connection.
+	AdditionalHeaders http.Header
+	// FetchURLTimeout is the timeout for the fetching the Quic url. If this is not set, the default is 30 seconds.
+	FetchURLTimeout time.Duration
+	// The client to used to connect to the redirect server
+	HttpClient arrangehttp.ClientConfig
```
schmidtw: Can we make the assumption that the redirect server will also be Quic & remove this field?

cmd/xmidt-agent/config.go
```diff
@@ -119,6 +121,34 @@ type Websocket struct {
...
+	// Config for quic connection
+	QuicClient myquic.Http3ClientConfig
+	// MaxMessageBytes is the largest allowable message to send or receive. (TODO
+	MaxMessageBytes int64
+	// (optional) Whether or not to use a redirect server like petasos or connect
+	// directly to a specific server.  Defaults to true.
+	UseRedirectServer bool
```
schmidtw: This should be transparent & automatically follow redirection.

cmd/xmidt-agent/config.go
```diff
@@ -119,6 +121,34 @@ type Websocket struct {
...
+	// RetryPolicy sets the retry policy factory used for delaying between retry attempts for reconnection.
+	RetryPolicy retry.Config
+	// Once sets whether or not to only attempt to connect once.
+	Once bool
```
schmidtw, line 146: Missing: `DisableV4` and `DisableV6`.

cmd/xmidt-agent/config.go
piccione99: it can't automatically follow redirection right now (there is no way I found to access the client connection after the redirect).  but I can intercept the response to see if it was redirected and then create a new connection.

cmd/xmidt-agent/config.go
piccione99: why do we need to explicitly disable v4 or v6?  Go automatically tries v6 first and then v4 if v6 fails.  I thought that's what we wanted?

cmd/xmidt-agent/config.go
piccione99: I'm not entirely sure what you mean by an arrangehttp style of help?    do you mean add http3 support to the arrange library?   I hope not?  (currently http client is not even used to connect via http3)

cmd/xmidt-agent/config.go
piccione99: And I know this sucks, but the other problem is I can't use an http client at all to connect to the server.  When I used an http client, it would create a new quic connection under the covers.  I'm using a different http3 api to connect so it actually uses the quic connection.  That's why I had the redirect flag.  I can still discover that programmatically instead, but it's not going to be pretty either way.

piccione99: FYI - this is the only way I found so far to get the api to use the quic connection I want it to (and this client does not redirect, so an http client has to be used to get the redirect):  

roundTripper := &http3.Transport{
		TLSClientConfig: qd.tlsConfig,
		QUICConfig:      &qd.quicConfig,
	}

h3Conn := roundTripper.NewClientConn(conn)

reqStream, err := h3Conn.OpenRequestStream(ctx)
if err != nil {
return nil, err
}

piccione99: I have another idea, hopefully nevermind all of above.

cmd/xmidt-agent/config.go
schmidtw: I think we will need to extend arrange ... to be able to control `tls.Config` correctly we will need it to work like arrange extends it.  This can be done after the merge.

cmd/xmidt-agent/config.go
schmidtw: It's fine if we have to do the extra work.  It should just be expected & not something we configure.

cmd/xmidt-agent/config.go
schmidtw: For now this is fine.  We will need to be able to disable v4 to prevent adding costs when an MSO doesn't own enough IPv4 addresses.  This flexibility also helps work around other networking conditions in testing/etc.

cmd/xmidt-agent/config.go
piccione99: I removed the redirect config and will retrieve the url from the redirect in a separate connection then create the new connection.  It still sucks, but there is no more config.

cmd/xmidt-agent/config.go
piccione99: done

cmd/xmidt-agent/config.go
piccione99: Just keep in mind, I'm not using http client from the client side, it seemed to always re-create the quic connection internally, no matter what.  I don't know how great that is going to work with arrange.

cmd/xmidt-agent/config.go
piccione99: ok, that should be automatically covered in the refactor, anyway

PR #602, opened 2026-07-28
fix: send the full device id with mac prefix in x-midt-mac-address header

ilawjr:
The fix changes the way the header is submitted.  It no longer strips the prefix, but sends it through instead.

Adds unit test.


internal/credentials/credentials_test.go
```diff
@@ -263,6 +263,50 @@ func TestEndToEnd429(t *testing.T) {
 	assert.Equal(1, called)
 }
 
+func TestFetchRequestHeaders(t *testing.T) {
```
denopink: we should modify an existing e2e test to perform this check, for example test `TestEndToEnd` could be a good home

denopink (changes requested):
a great start 🍻

denopink (approved):
I completed the rest of work required for this pr.

denopink:
closing since some commits are unverified

